PR #48, opened 2021-11-14
feat: create heading components

supergrecko:
These are already styled from typography.ts


kharann:
where's my heading stories

PR #50, opened 2021-11-14
Curves n circles

NjaalSoerland:
Created a molecule with Circles connected by dotted lines based on the figma file. Both the Circle atom and ConnectedCircles molecule is scalable through the size parameter


src/components/atoms/stories/Circle.stories.tsx
```diff
@@ -0,0 +1,10 @@
+import { Circle } from "../Circle";
+
+export default {
+    title: "atoms/Circle",
+    component: Circle,
+};
+
+export const Primary = () => <Circle size={50}>1</Circle>;
```
kharann, line 8: Call this medium

kharann (changes requested):
add a prop to hange color

kharann (approved):
lgtmgltm

PR #66, opened 2022-02-16
Convert company page to stitches and refactor it

kharann:
- Convert styling to use stitches in company page
- Merge the components that simply forwards a PortableText component
- Move all styling to a styles object at the bottom


Thomhas (approved):
Looks good! You never fail to amaze! 😄

PR #71, opened 2022-02-27
Badge stitches

sondrealf:


src/components/atoms/Badge.tsx
```diff
@@ -0,0 +1,52 @@
...
+    },
+  },
+});
+
+const Badge = styled("p", {
+  borderRadius: "5px",
+  fontWeight: "600",
+  padding: "0 0.5rem",
+  display: "inline-block",
+  variants: getVariant(),
```
kharann, line 49: I think you should put all variants directly as an object rather than a function

src/components/atoms/Badge.tsx
```diff
@@ -0,0 +1,52 @@
...
+      borderColor: "$$main",
+    },
+    outline: {
+      color: "$$main",
+      bg: "$$secondary",
+    },
+  },
+});
+
+const Badge = styled("p", {
```
kharann, line 44: Should be span rather than p. P should be only used for paragraphs.

src/components/atoms/Badge.tsx
```diff
@@ -0,0 +1,52 @@
...
+  borderRadius: "5px",
```
kharann, line 45: use $2 rather than 5px

src/components/atoms/Badge.tsx
```diff
@@ -0,0 +1,52 @@
...
+  },
+  variant: {
+    solid: {
+      color: "$white",
+      bg: "$$main",
+    },
+    subtle: {
+      bg: "none",
+      color: "$$main",
+      border: "1px solid",
```
kharann, line 34: can maybe be written as `border: "1px solid $$main"`

kharann (changes requested):

kharann (approved):
lgtm

PR #117, opened 2022-07-09
:lipstick: #115 add alert modal component

### LINKED ISSUE #115: Component: Modal dialog

VigdisIS:
Closes #115

### DIFF
--- a/packages/ow-ui/src/components/Button/Button.tsx
+++ b/packages/ow-ui/src/components/Button/Button.tsx
@@ -22,7 +22,7 @@ const Button = styled("button", {
     backgroundColor: "$blue7",
   },
   variants: {
-    textOnly: {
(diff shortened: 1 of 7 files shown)


packages/ow-ui/src/components/Button/Button.tsx
```diff
@@ -22,6 +22,14 @@ const Button = styled("button", {
     backgroundColor: "$blue7",
   },
   variants: {
+    textOnly: {
```
kharann, line 25: call this subtle perhaps

kharann (approved):

packages/ow-ui/src/components/Button/Button.tsx
VigdisIS: fixdd

PR #121, opened 2022-07-10
split alerticon to its own internal component

kharann:


packages/ow-ui/src/components/Alert/AlertIcon.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+interface AlertIconProps {
+  status: "info" | "warning" | "success" | "danger"
+  className?: string
+}
+
+export const AlertIcon: FC<AlertIconProps> = ({ status, className }) => {
+  const style = `${styles.base({ status })} ${className}`
+  switch (status) {
+    case "info":
```
Thomhas, line 13: The alert icon needs a white variant for when it is used on top of color!

Thomhas (changes requested):

packages/ow-ui/src/components/Alert/AlertIcon.tsx
Thomhas: The code looks really nice!

Issue #108, opened 2022-07-09
Component: Flag


VigdisIS:
Ref flag from ds figma

Thomhas:
Flag and toast are now the same component ref: #132

PR #127, opened 2022-07-10
Feat/flag ds

Thomhas:


Thomhas:
Closes: #108

matsjla (changes requested):
Looks very nice! :+1: two tiny nits that should be fixed: 

The collapsable cards are missing border-radius in the bottom corners when collapsed
image

The "Understood" text isn't aligned with the rest of the card
image

kharann (changes requested):
dependency in wrong project, should be in packages/ow-ui/package.json

Issue #126, opened 2022-07-10
Component: Dropdown Menu


matsjla:
Follow Figma design

PR #129, opened 2022-07-10
feat: DropdownMenu with basic text options and loading

matsjla:
Related to #126 

An initial version of the component with basic text options available and a loading flag+indicator


Thomhas (commented):
Really good job!

PR #175, opened 2022-10-05
Company users

NjaalSoerland:
Closes: #

## Checklist

- [X] I have written tests

## Changelog

- Updated the company class
- Added setters and getters for company objects
- Added tests

## How to test

Run yarn test (woah!)


kharann:
Perhaps the company type should be an ENUM tho. Since there's not that  many different types

PR #198, opened 2022-10-26
Config overhaul

kharann:
## Changelog
- Move config to a /config folder
- Set a config for eslint and prettier
- Add a import sort order plugin for prettier

Go to [config](./packages/config) to see the current configs.
The large amount of changes is due to the import sort order plugin


matsjla:
Never used prettier plugins before, but there is a very nice import order plugin for Eslint which lets you customize A LOT of stuff like ordering, groups, etc.

kharann:
> Never used prettier plugins before, but there is a very nice import order plugin for Eslint which lets you customize A LOT of stuff like ordering, groups, etc.

Decided to use prettier since it will automatically sort the orders on  format

kharann:
@matsjla BUMP

PR #222, opened 2022-11-08
Fix type errors from backend and prisma

kharann:
## Change log
- export a global default Prisma
- instead of using models, we use the type `Prisma.___GetPayload<null>`. That way we can avoid conflicts between the typing of prisma and the normal types
- owdb is copied from cal.com

### DIFF
--- a/apps/ow-api/src/context.ts
+++ b/apps/ow-api/src/context.ts
@@ -1,24 +1,21 @@
-import { PrismaClient } from "@dotkomonline/db"
-import { type Session } from "next-auth"
+import { prisma } from "@dotkomonline/db"
 
-import { type inferAsyncReturnType } from "@trpc/server"
+import type { inferAsyncReturnType } from "@trpc/server"
+import * as trpcExpress from "@trpc/server/adapters/express"
 
 import { initUserRepository } from "./modules/auth/user-repository"
 import { initUserService } from "./modules/auth/user-service"
 import { initEventRepository } from "./modules/event/event-repository"
 import { initEventService } from "./modules/event/event-service"
 
-type CreateContextOptions = {
-  session: Session | null
-}
-
 /** Use this helper for:
  * - testing, so we dont have to mock Next.js' req/res
  * - trpc's `createSSGHelpers` where we don't have req/res
  * @see https://beta.create.t3.gg/en/usage/trpc#-servertrpccontextts
  **/
-export const createContext = async (_opts: CreateContextOptions) => {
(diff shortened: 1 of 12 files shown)


apps/ow-api/src/context.ts
```diff
@@ -1,24 +1,21 @@
...
 /** Use this helper for:
  * - testing, so we dont have to mock Next.js' req/res
  * - trpc's `createSSGHelpers` where we don't have req/res
  * @see https://beta.create.t3.gg/en/usage/trpc#-servertrpccontextts
  **/
-export const createContext = async (_opts: CreateContextOptions) => {
-  const client = new PrismaClient()
+export const createContext = async ({ req, res }: trpcExpress.CreateExpressContextOptions) => {
+  console.log({ req, res })
+  const client = prisma
```
matsjla, line 18: maybe just delete variable?

packages/ow-db/src/index.ts
```diff
@@ -1,2 +1,20 @@
...
+export const prisma = globalThis.prisma || new PrismaClient(prismaOptions)
+
+export const customPrisma = (options: Prisma.PrismaClientOptions) => new PrismaClient({ ...prismaOptions, ...options })
+
+if (process.env.NODE_ENV !== "production") {
+  globalThis.prisma = prisma
+}
+
+export type { Prisma, PrismaClient }
+export default prisma
```
matsjla, line 20: What is the rationale behind making it available as global variable? I'm guessing some SSR thing with Next maybe but?

Also maybe remove either default or named export so there's not two?

kharann: Personally, I just looked at the example in [turobopack](https://github.com/vercel/turbo/blob/main/examples/with-prisma/packages/database/src/client.ts) example repo, and what is done in calcom. Looks like this is about ensuring there's a single instance of prisma in the application (Singleton pattern).

kharann: I would guess that it is to avoid multiple connections perhaps from the same server

matsjla: I *don't think* that would matter since Node will cache the module ensuring only one prisma instance is created but again, unsure how that works with bunding/next ssr/etc

matsjla: Probably fine to leave it as-is though LGTM

kharann: As you mentioned, it might be for Nextjs. I'm considering using prisma for NextAuth and auth session handling (kinda prefer redis, but well well).

PR #229, opened 2022-11-09
refactor: merge auth and event trpc routers

matsjla:
Moves the signing mutation into the auth namespace and merges with the event router


kharann (approved):
lgtm

PR #313, opened 2023-02-07
Path alias

kharann:
## Changelog

- Remove route aliases like `@components/` and replace all of it with `@/components/`
- Remove vitest tsconfig paths


matsjla (approved):
lgtm

PR #318, opened 2023-02-15
Added an editevent method to eventrepository

jotjern:


matsjla (approved):
Looks great! Just need to fix the lint errors are we're good to go :rocket:

PR #329, opened 2023-02-22
Added a list users endpoint to the auth router

jotjern:


packages/api/src/modules/auth/auth-router.ts
```diff
@@ -27,4 +27,7 @@ export const authRouter = t.router({
   skipLogin: t.procedure.input(z.object({ challenge: z.string() })).mutation(({ input, ctx }) => {
     return ctx.userService.skipLogin(input.challenge)
   }),
+  getUsers: t.procedure.input(z.object({ limit: z.number().optional() })).query(async ({ input, ctx }) => {
+    return await ctx.userService.getUsers(input.limit ?? 10)
```
matsjla, line 31: Can we make the default 50?

matsjla (approved):

PR #336, opened 2023-02-28
Add form builder abstraction

matsjla:
Allows you to share form templates and create custom inputs. Extremely flexible because all props can be passed down to component.

Also makes the user/committee pages look the same as the event one


kharann (approved):
nice

PR #338, opened 2023-03-03
Add pagination and refactor services into classes

kharann:
## Changelog

- More fluent way to paginate lists of data
- Fixed broken vitests
- Converted companies to the new class service format

### DIFF
--- a/apps/dashboard/src/app/event/EventCreationModal.tsx
+++ b/apps/dashboard/src/app/event/EventCreationModal.tsx
@@ -10,7 +10,7 @@ export type EventCreationModalProps = {
 }
 
 export const EventCreationModal: FC<EventCreationModalProps> = ({ close }) => {
-  const { data: committees = [] } = trpc.event.all.useQuery()
+  const { data: committees = [] } = trpc.committee.all.useQuery()
   const utils = trpc.useContext()
   const create = trpc.event.create.useMutation({
     onSuccess: () => {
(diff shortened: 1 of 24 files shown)


apps/dashboard/src/app/event/EventCreationModal.tsx
```diff
@@ -10,7 +10,7 @@ export type EventCreationModalProps = {
 }
 
 export const EventCreationModal: FC<EventCreationModalProps> = ({ close }) => {
-  const { data: committees = [] } = trpc.committee.all.useQuery({ limit: 50 })
+  const { data: committees = [] } = trpc.event.all.useQuery()
```
matsjla, line 13: Should be trpc.committee.all

packages/api/src/utils/db-utils.ts
```diff
@@ -0,0 +1,25 @@
...
+    cursor: CursorSchema.optional(),
+  })
+  .optional()
+  .default({ take: 20, cursor: undefined })
+
+export type Cursor = z.infer<typeof CursorSchema>
+
+export const paginateQuery = (qb: AnySelectQueryBuilder, cursor: Cursor) => {
+  return qb
+    .where((qb) => qb.where("createdAt", "=", cursor.createdAt).where("id", "<", cursor.id))
```
matsjla, line 21: How does this work? Doesn't this become uuid comparison?

matsjla (changes requested):

packages/api/src/utils/db-utils.ts
kharann: I just did this https://dba.stackexchange.com/questions/267794/how-to-do-pagination-with-uuid-v4-and-created-time-on-concurrent-inserted-data

kharann: image

matsjla (approved):
epic

PR #365, opened 2023-04-26
fix: re-export cockroach dialect and adapter separately from migrator

matsjla:
Addresses a problem where the migrator wouldn't run because the adapter wasn't exported, while still making it possible for nextjs to run without having to resolve the new URL in createMigrator


kharann:
We dont really use the cockroachdb migrator anymore as we have moved to use Postgres dialect

PR #375, opened 2023-04-29
Add payment system using stripe

Terbau:
Adds payment integration using Stripe Checkout.

Quick functionality overview:
- Add and manage products
- Create checkout sessions if allowed by the product
- Link other objects (e.g. events) to products

Some additional notes:
- Multiple stripe keys are supported
- Integration with other payment services (e.g. vipps) was taken into account and should be easy to implement in the future

## Checklist

- [ x] I have written tests
- [ x] I have provided documentation

## How to test

See the added part of the readme :)


apps/web/src/pages/api/payment/stripe/[publickey].ts
```diff
@@ -0,0 +1,77 @@
...
+  if (!stripe || !endpointSecret) {
+    console.warn("No stripe account found for the given public key")
+    res.status(400).end()
+    return
+  }
+
+  let event: Stripe.Event
+  try {
+    const body = await buffer(req)
+    event = stripe.webhooks.constructEvent(body, sig, endpointSecret)
```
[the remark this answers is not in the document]
Comment (Terbau) [reply apps/web/src/pages/api/payment/stripe/[publickey].ts:28]: false positive, ignore

apps/web/src/pages/api/payment/stripe/[publickey].ts
```diff
@@ -0,0 +1,77 @@
+import { NextApiRequest, NextApiResponse } from "next"
+import { getStripeObject, getStripeWebhookSecret } from "@dotkomonline/api/src/lib/stripe"
```
Comment (kharann) [inline apps/web/src/pages/api/payment/stripe/[publickey].ts:2]: should probably export the whole handler in @dotkomonline/api and export it. Then you should just import the handler and do

```js
import { paymentHandler } from "@dotkomonline/api"

export default paymentHandler
```

apps/web/src/pages/api/payment/stripe/[publickey].ts
```diff
@@ -0,0 +1,77 @@
...
+    const body = await buffer(req)
+    event = stripe.webhooks.constructEvent(body, sig, endpointSecret)
+  } catch (e) {
+    const err = e as Error
+    res.status(400).send(`Webhook Error: ${err.message}`)
+    return
+  }
+
+  if (event.type.startsWith("checkout.session")) {
+    const ctx = await createContextInner({ auth: null })
```
Comment (kharann) [inline apps/web/src/pages/api/payment/stripe/[publickey].ts:36]: huh how do you have access to this in the web folder? If this handler is moved to the API folder, you don't need to import this

kharann (changes requested):

packages/api/src/lib/stripe.ts
```diff
@@ -0,0 +1,49 @@
+import Stripe from "stripe"
+
+interface StripeAccountDetails {
+  publicKey: string
+  secretKey: string
+  webhookSecret: string
+}
+
+const stripeAccountsMap = new Map<string, StripeAccountDetails>()
```
kharann, line 9: i think this is a bit sus. Maps are usually used when the quantity of elements are unknown. In this case, the amount of possible stripe keys exists. You should make an object, with the `as const` to ensure that the object is readonly, and therefore each key is fully autocompletable

kharann (changes requested):

packages/api/src/modules/payment/payment-service.ts
```diff
@@ -0,0 +1,16 @@
+import { PaymentProvider } from "@dotkomonline/types"
+import { readableStripeAccounts } from "./../../lib/stripe"
+
+export interface PaymentService {
+  getPaymentProviders(): (PaymentProvider & { paymentAlias: string })[]
+}
+
+export class PaymentServiceImpl implements PaymentService {
```
kharann, line 8: This class seems a bit small? mayhaps it should be merged with another service

packages/types/src/payment.ts
```diff
@@ -0,0 +1,63 @@
...
+  type: z.enum(["EVENT"]), // inlude WEBSHOP later
+  objectId: z.string().uuid().nullable(), // The OW object this product is linked to e.g. eventId, webshopItemId. null if not linked
+  amount: z.number(), // price
+  paymentProviders: z.array(PaymentProviderSchema),
+  deletedAt: z.date().nullable(), // null = not deleted
+})
+
+export type Product = z.infer<typeof ProductSchema>
+
+export const ProductWriteSchema = ProductSchema.partial({
```
kharann, line 34: Probably should be omit, as these should never be able to be updateable,

apps/web/src/pages/api/payment/stripe/[publickey].ts
Comment (Terbau) [reply apps/web/src/pages/api/payment/stripe/[publickey].ts]: Fixed

packages/api/src/lib/stripe.ts
Terbau: I agree. Fixed.

packages/types/src/payment.ts
Terbau: Fixed

packages/api/src/modules/payment/payment-service.ts
Terbau: Agreed. Everything named transaction is now named payment, and the old transaction service and router was merged with the existing ones. The product router is still accessible as a subrouter through payment router.

matsjla (approved):

PR #376, opened 2023-04-30
Add a clerk webhook to add the user ID to local database on registration

kharann:
## Changelog

- This webhook adds the user to our database when the event, `user.created` is sent.
- Removes all unecesarry fields from the user object
- Consider renaming user to Tenant?
- Rewrites some old migrations, so probably need to fix the migrations


Terbau (approved):
LGTM

PR #389, opened 2023-05-09
Add committee page

Terbau:
## Changelog

- Adds a page to display info about a committee. Uses the same layout as the company page
- Extends the committee table with additional fields such as `description`, `email` and `image`

## How to test

1. Run migrations (I recommend `pnpm migrate-with-fixtures` for the latest fixture data)
2. Spin monoweb up and go to `/committee`

## Preview
image

### DIFF
--- a/apps/sanity/package.json
+++ b/apps/sanity/package.json
@@ -19,7 +19,8 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-is": "^18.2.0",
-    "sanity": "^3.0.0"
+    "sanity": "^3.0.0",
+    "styled-components": "^5.2"
   },
   "devDependencies": {
     "@sanity/eslint-config-studio": "^2.0.1",
--- a/apps/web/src/components/layout/EntryDetailLayout.tsx
+++ b/apps/web/src/components/layout/EntryDetailLayout.tsx
@@ -1,0 +1,25 @@
+import { FC, PropsWithChildren } from "react"
+import { cn } from "@dotkomonline/ui"
+
+export interface EntryDetailLayoutProps {
+  title: string
+  type?: string
+  color?: "BLUE" | "GREEN" | "AMBER"
+}
+
+export const EntryDetailLayout: FC<PropsWithChildren<EntryDetailLayoutProps>> = ({ children, title, type, color }) => {
+  const borderColorClass = color ? `border-${color.toLowerCase()}-8` : "border-blue-8"
+  const textColorClass = color ? `text-${color.toLowerCase()}-11` : "text-blue-11"
+
+  return (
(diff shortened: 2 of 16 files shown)


apps/web/src/components/layout/EntryDetailLayout.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export interface EntryDetailLayoutProps {
+  title: string
+  type?: string
+  color?: "BLUE" | "GREEN" | "AMBER"
+}
+
+export const EntryDetailLayout: FC<PropsWithChildren<EntryDetailLayoutProps>> = ({ children, title, type, color }) => {
+  const borderColorClass = color ? `border-${color.toLowerCase()}-8` : "border-blue-8"
+  const textColorClass = color ? `text-${color.toLowerCase()}-11` : "text-blue-11"
```
matsjla, line 12: I don't think tailwind is clever enough to strip this at compile time, should do something like
```suggestion
export const EntryDetailLayout: FC<PropsWithChildren<EntryDetailLayoutProps>> = ({ children, title, type, color = 'BLUE' }) => {
  const borderColorClass = cn({
    "border-blue-8": color === 'BLUE',
    "border-green-8": color === 'GREEN',
    "border-amber-8": color === 'AMBER',
  })
  // same for textColorClass
```
})

packages/config/tailwind-preset.js
```diff
@@ -82,6 +82,14 @@ module.exports = {
...
     require("tailwindcss-animate"),
   ],
+  safelist: [
+    {
+      pattern: /text-(blue|amber|red|green|indigo)-11/,
+    },
+    {
+      pattern: /border-(blue|amber|red|green|indigo)-8/,
+    },
+  ],
```
matsjla, line 92: I think it's better to describe all the classnames that will be used than maintaining the list here

apps/sanity/package.json
```diff
@@ -19,7 +19,8 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-is": "^18.2.0",
-    "sanity": "^3.0.0"
+    "sanity": "^3.0.0",
+    "styled-components": "^5.2"
```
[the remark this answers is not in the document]
Terbau: Can't be bothered to fix this anymore. Problem for a another time Cx

apps/web/src/components/layout/EntryDetailLayout.tsx
Terbau: In my experience this should work, but I agree. Its harder to maintain without knowing exactly how tailwind works, and its way more clear to use your suggested and more intuitive method. I'll get it sorted.

PR #396, opened 2023-05-15
Fix migrations

Terbau:
Fixes some issues with the previous two migrations written that resulted in the migration process failing.

### DIFF
--- a/packages/ow-db/src/migrations/0011_alter_committee.ts
+++ b/packages/ow-db/src/migrations/0011_alter_committee.ts
@@ -4,7 +4,7 @@ export async function up(db: Kysely<any>) {
   await db.schema
     .alterTable("committee")
     .addColumn("description", "text", (col) => col.notNull().defaultTo(""))
-    .addColumn("email", "text", (col) => col.notNull())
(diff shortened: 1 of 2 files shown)


packages/ow-db/src/migrations/0011_alter_committee.ts
```diff
@@ -4,7 +4,7 @@ export async function up(db: Kysely<any>) {
   await db.schema
     .alterTable("committee")
     .addColumn("description", "text", (col) => col.notNull().defaultTo(""))
-    .addColumn("email", "text", (col) => col.notNull())
+    .addColumn("email", "text", (col) => col.notNull().defaultTo(""))
```
matsjla, line 7: Perhaps make it default to some online.ntnu.no email?

matsjla (approved):

packages/ow-db/src/migrations/0011_alter_committee.ts
Terbau: Yeah, thats probably a good idea. Maybe `kontakt@online.ntnu.no`?

matsjla: sounds good

PR #404, opened 2023-06-16
Add privacy permissions

Terbau:
- Adds privacy permissions for users to the backend. Fields are as follows:
  - `profileVisible` (default: `true`)
  - `usernameVisible` (default: `true`)
  - `emailVisible` (default: `false`)
  - `phoneVisible` (default: `false`)
  - `addressVisible` (default: `false`)
  - `attendanceVisible` (default: `false`)

## Checklist

- [x] I have written tests


kharann:
look at him work

kharann:
OG MED TESTER

matsjla (approved):
:rocket:

kharann:
Although most of the content is satisfactory, it remains uncertain whether we require all of the changes. It primarily depends on whether we plan to still implement the profile search and which data we will still store.

One important consideration is placing the privacy permission within the user table instead of a separate table. This change would simplify the process by reducing the number of joins required in certain scenarios, given that the user data is already stored in an external service like Clerk. Overall this change would perhaps allow the User table to be used for something, rather than it being a table with a single ID column.

PR #411, opened 2023-08-01
Split packages/api into gateway-edge-next, gateway-trpc and core

matsjla:
Splits the packages/api module into three new modules:

1. core: entire application service layer (services, repositores)
2. gateway-trpc: trpc context/router to serve portions of the service layer through trpc
3. gateway-edge-next: edge functions to be made available through next api routes. This is called edge-next because the functions have a hard dependency on nextjs' api

Rationale behind this change is that the service layer should not have to care about what gateway is serving it. Splitting the packages up like this means no gateway dependencies (trpc, next) make it to the service layer

Also changes makes package naming convention consistent (remove ow- prefix)


Terbau (approved):
Runs find and looks good to me. Also like the philosophy behind the refactor :)

PR #412, opened 2023-08-01
Remove dependency on clerk/nextjs sdk in non-next components

matsjla:
Makes use of the default node sdk for components that dont depend on nextjs


Terbau (approved):
LGTM

PR #432, opened 2023-08-18
Use AWS Cognito for user management

matsjla:
Configures an AWS Cognito user pool with two app clients (dashboard, web) and uses Next Auth with CognitoProvider to manage authentication and sessions.

Nukes everything referencing Hydra and Clerk as they are not needed anymore.

## Future work

- Create custom login/signup pages, right now we have AWS default
- Create Cognito workflow trigger on user signup to create database user
- Create service methods for finding and querying AWS Cognito users


matsjla:
Vercel fails bc environment variables aren't updated there

Terbau (approved):
2 hrs later... lgtm sxy mf

PR #485, opened 2023-09-25
Switch to ulid

junlarsen:
Makes it so we dont have to mental gymnastics to ensure that ordering on createdAt is consistent. Makes cursors simpler

## Why ulid?

Because the generated ids are monotonically sortable (K-sortable). This means we can sort them. The possible security concern of leaking the creation order of items doesn't really matter for our cases.

Interesting reads:
- https://github.com/paralleldrive/cuid2?tab=readme-ov-file#the-contenders
- https://supabase.com/blog/choosing-a-postgres-primary-key


junlarsen:
Build fails because dev database isn't re-migrated

junlarsen:
Blocked by https://community.neon.tech/t/support-for-v0-1-3-of-pgx-ulid/1194

henrikskog:
suiii lets go

henrikskog (approved):
suiii lets go 🚀

PR #536, opened 2023-09-29
fix: consistently name all apps

junlarsen:
All packages were good from before


kharann:
bump commit i guess to check for pipeline

PR #538, opened 2023-09-29
refactor: migrate to terraform modules

junlarsen:
Makes proper use of Terraform modules to provision infra. Monoweb infra started to reach a scale where managing many small terraform projects became annoying.

Refactored into modules (aws-acm-certificate, aws-cognito-pool-client, etc) which are consumed by a root project.

Infra was also redeployed


kharann (approved):
use opentofu or you're anti open-source (jk)

PR #543, opened 2023-10-02
refactor: restructre emails to work with email dev

junlarsen:
Does some minor changes in order to make `email dev` compatible with our email setup.

Also temporarily removes the email handler (to be moved to lambda)


kharann (approved):
lgtm

PR #562, opened 2023-10-04
Feature/dot 162 student year progress component

norawl:
This component displays which year the student is currently in on the Profile landing page.
NB: Currently doesn't retrieve any data, the year is set manually to 0.


henrikskog:
Niiiice! I noticed that in light mode the text inside the circles have quite bad contrast to the background. Maybe use a lighter font for light mode? 
image

PR #570, opened 2023-10-05
refactor: expand eslint config

brage-andreas:
* Expand the ESLint config
* Expand prettier config
* Fix (most) created problems

---

Why?
* By narrowing down your options you prevent bugs
* The code becomes easier to read due to uniformity and added whitespace

Possible problems:
* The config as-is is highly opinionated
* Some rules may seem annoying (or difficult) to deal with. We will have to see!


packages/config/eslint-preset.js
```diff
@@ -1,40 +1,148 @@
...
-      extends: ["plugin:@typescript-eslint/recommended", "plugin:@typescript-eslint/eslint-recommended"],
-      plugins: ["@typescript-eslint"],
-      parser: "@typescript-eslint/parser",
-      rules: {
-        "@typescript-eslint/no-empty-interface": 0,
-      },
+    env: {
+        browser: true,
+        es6: true,
+        node: true,
```
junlarsen, line 10: This isn't necessarily true so it shouldn't be in the base config

packages/config/eslint-preset.js
```diff
@@ -1,40 +1,148 @@
...
+        "plugin:perfectionist/recommended-natural",
+        "plugin:react/recommended",
+        "plugin:react-hooks/recommended",
+        "plugin:jsx-a11y/recommended",
+        "turbo",
+        "prettier",
+    ],
+
+    globals: {
+        React: "writable",
```
junlarsen, line 26: We dont have a React global in React 18+ when specifying jsx: react-jsx

packages/config/eslint-preset.js
```diff
@@ -1,40 +1,148 @@
...
+            {
+                ObjectExpression: {
+                    consistent: true,
+                    minProperties: 4,
+                    multiline: true,
+                },
+            },
+        ],
+        "object-shorthand": "error",
+        "padding-line-between-statements": [
```
junlarsen, line 107: Personally not a fan of this, leads to a lot of empty lines in code that can be very concise otherwise

packages/config/eslint-preset.js
```diff
@@ -1,40 +1,148 @@
...
+        project: [path.resolve(__dirname, "tsconfig.json"), "./packages/*/tsconfig.json"],
+        sourceType: "module",
+        tsconfigRootDir: __dirname,
+    },
+
+    plugins: ["@typescript-eslint"],
+
+    root: true,
+
+    rules: {
```
junlarsen, line 51: Could you explain the rules added, and the value of each of them? I fear we might be adding a lot of restrictions for little gain

.prettierrc
```diff
@@ -1,9 +1,9 @@
 {
-  "semi": false,
-  "trailingComma": "es5",
-  "singleQuote": false,
-  "printWidth": 120,
-  "endOfLine": "auto",
-  "arrowParens": "always",
-  "plugins": ["prettier-plugin-tailwindcss"]
+    "printWidth": 120,
+    "tabWidth": 4,
```
junlarsen, line 3: I personally find tabWidth 4 to become very annoying with JSX because of the heavy indentation.

I also don't think that it's worth re-formatting every part of the codebase for this.

.prettierrc
```diff
@@ -1,9 +1,9 @@
...
+    "quoteProps": "consistent",
+    "trailingComma": "es5",
```
junlarsen, line 5: I think this one is nice to have "all" for, makes diffs smaller which is great for review

.prettierrc
```diff
@@ -1,9 +1,9 @@
 {
-  "semi": false,
```
junlarsen, line 2: Not too opinionated on this, but I don't see a good reason to change it.

junlarsen (changes requested):
Interesting! I think there are a lot of nice improvements here, but also quite a few rules that we should discuss the value of before adding.

.prettierrc
henrikskog: boo

PR #586, opened 2023-10-07
DOT-310 Replace storybook with ladle

kharann:
idk maybe storybook is just too big


junlarsen (commented):
Interesting, what are the benefits/drawbacks vs storybook?

junlarsen:
oh its kinda hot https://ladle.dev/docs/

kharann:
> Interesting, what are the benefits/drawbacks vs storybook?

like 20x smaller. fast af

image

we don't really use any of these features

kharann:
We could eventually consider having the stories in a stories folder instead of inside each package in UI :thinking:

junlarsen (approved):

PR #587, opened 2023-10-09
Add logic to prevent non logical start and end dates

henrikskog:
**Create event modal**
- Don't allow start time or end time in the past
- Don't allow start time that's set before end time

**Event edit page**
- Don't allow start-time before end-time

**Attendance**
- I have not made changes to attendance stuff here as this is not a part of the ui yet anyway

**For context to understanding changes**
I split create schema and edit schema up to allow for different validation logic in create and edit forms. (see above). Also, when using refine/transforms with zod, a ZodEffect container type is returned. I did not find a way to use .required() on that without loosing the refinement. Therefore, this split was needed anyway.

Also, you could reduce code duplication a lot by refactoring the form components into a component with the fields, then have the schema passed in or something. I tried and it required some ts magic so I dont know, if someone wanna help then we can do that


junlarsen (commented):
I think it's better to just keep the same WriteForm and instead implement it at the service level, do a check in EventService and see if the dates are valid

ZodEffects are also kind of annoying to deal with too, and it likely makes it more trouble than what it's worth

henrikskog:
I agree that this got a bit messy. In your approach, are you thinking of just showing the error in a notification? I have implemented another approach validating client side. Passing the setError from hookform's useForm into the onSubmit function which allows for custom validation there. Not sure about that approach either, but I think its very nice to get the feedback under the appropriate input element

kharann:
Is there a reason for not using

`z.date().max(new Date(), { message: "Ikke I framtiden?!" });`

henrikskog:
Nope, that's cleaner!

apps/dashboard/src/app/(dashboard)/event/[id]/edit-card.tsx
```diff
@@ -1,17 +1,20 @@
...
 export const EventEditCard: FC = () => {
   const { event } = useEventDetailsContext()
   const edit = useEditEventMutation()
-  const FormComponent = useEventWriteForm({
+  const FormComponent = useEventEditForm({
     label: "Oppdater arrangement",
     onSubmit: (data) => {
-      const result = EventWriteSchema.required({ id: true }).parse(data)
-      edit.mutate(result)
+      const result = EventEditSchema.parse(data)
```
Comment (kharann) [inline apps/dashboard/src/app/(dashboard)/event/[id]/edit-card.tsx:13]: I agree with Mats that there should be a check in EventService to ensure the incoming data is valid and throw and error if not. However, you could also add an check in the submit handler which sets the field to dirty on error.

packages/db/src/migrations/0016_autogenerate_user_id.ts
```diff
@@ -8,5 +8,8 @@ export async function up(db: Kysely<any>) {
 }
 
 export async function down(db: Kysely<any>) {
-  await db.schema.alterTable("ow_user").alterColumn("id", (col) => col.setDefault(sql`""`))
+  await db.schema
+    .alterTable("ow_user")
+    .alterColumn("id", (col) => col.setDefault(sql`null`))
```
henrikskog, line 13: this is just a hotfix for the sql`""` which is invalid

henrikskog:
I don't really see the point in enforcing this stuff in the event service as its mostly about helping people not mis-input the dates and not business critical stuff. But let me know if you think otherwise!

packages/db/src/migrations/0016_autogenerate_user_id.ts
junlarsen: You can delete this migration, it's no longer in the main branch

junlarsen (approved):

PR #595, opened 2023-10-11
refactor: expand eslint and prettier config

brage-andreas:
* Expand the ESLint config
* Tidy the prettier config
* Fix most created problems


brage-andreas:
There are some React/JSX errors left that I did not touch.

packages/config/eslint-preset.js
```diff
@@ -1,40 +1,131 @@
...
+        prev: "import",
+      },
+    ],
+    "prefer-arrow-callback": "error",
+    "prefer-const": "error",
+    "prefer-template": "error",
+    "quote-props": ["error", "consistent-as-needed"],
+    "react/prop-types": "off",
+    "react/react-in-jsx-scope": "off",
+    "react/display-name": "off",
```
junlarsen, line 122: This one should not be disabled, it's very useful to have displayNmae when you are debugging forwardRefs etc

junlarsen (dismissed):

packages/config/eslint-preset.js
henrikskog: Take this component: packages/ui/src/components/Icon/Icon.tsx
```js
import { Icon as Iconify } from "@iconify-icon/react"
import { forwardRef } from "react"

export type IconProps = React.ComponentPropsWithoutRef<typeof Iconify>

export const Icon = forwardRef<React.ElementRef<typeof Iconify>, IconProps>((props, ref) => (
  <Iconify ref={ref} {...props} />
))
```

This current implementation has no display name. I understand this makes it hard to debug.

I first thought of fixing it like this:
```js
export const Icon = forwardRef<React.ElementRef<typeof Iconify>, IconProps>(function Icon(props, ref) {
  return <Iconify ref={ref} {...props} />
})
```

However this conflicts with the "prefer-arrow-callback" rule. So doing it like this we should probably disable that rule.

I guess option 2 would be to enforce this pattern? If we want to keep "prefer-arrow-callback"

```js
import { Icon as Iconify } from "@iconify-icon/react"
import { forwardRef } from "react"

export type IconProps = React.ComponentPropsWithoutRef<typeof Iconify>

function IconFn(props: IconProps, ref: React.Ref<React.ElementRef<typeof Iconify>>) {
  return <Iconify ref={ref} {...props} />
}

export const Icon = forwardRef<React.ElementRef<typeof Iconify>, IconProps>(IconFn)
```

junlarsen: The recommended way of doing it is with a .displayName accessor

```ts
import { Icon as Iconify } from "@iconify-icon/react"
import { forwardRef } from "react"

export type IconProps = React.ComponentPropsWithoutRef<typeof Iconify>

export const Icon = forwardRef<React.ElementRef<typeof Iconify>, IconProps>((props, ref) => (
  <Iconify ref={ref} {...props} />
))

Icon.displayName = 'Icon'
```

henrikskog: Aaa Nice!

henrikskog: Fixed in new commits

brage-andreas:
@junlarsen there are five `@typescript-eslint/no-unnecessary-condition` hits left.
I'm not confident I know how to handle them, as simply removing the "unnecessary" condition may lead to bugs I am unaware of.

Otherwise, this PR seems ready for review.

henrikskog (approved):
lets go boys

PR #597, opened 2023-10-12
keep expanded navbar labels when navigating dashboard

NjaalSoerland:
Stores the opened labels in local storage and reopens them on page load so that their state persists

### DIFF
--- a/apps/dashboard/src/app/ApplicationShell.tsx
+++ b/apps/dashboard/src/app/ApplicationShell.tsx
@@ -1,6 +1,6 @@
 "use client"
 
-import { FC, PropsWithChildren } from "react"
+import { FC, PropsWithChildren, useState, useEffect } from "react"
 import { useDisclosure } from "@mantine/hooks"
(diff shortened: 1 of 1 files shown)


apps/dashboard/src/app/ApplicationShell.tsx
```diff
@@ -66,7 +66,22 @@ const navigations = [
 export const ApplicationShell: FC<PropsWithChildren> = ({ children }) => {
   const [mobileOpened, { toggle: toggleMobile }] = useDisclosure()
   const [desktopOpened, { toggle: toggleDesktop }] = useDisclosure(true)
+  const [expandedNavs, setExpandedNavs] = useState<string[]>([])
 
+  useEffect(() => {
```
junlarsen, line 71: Replace useEffect with initializer in useState()

junlarsen (changes requested):

junlarsen (approved):

PR #599, opened 2023-10-13
Create event dashboard for adding secondary committees

henrikskog:
## What's changed
We want to support multiple committees organizing an event together. Before, event had a foreign key committeeId to the committe organizing the event. Now there is a many to many relation between event and committee.

The UI is changed to support this change.

## Before
image
image

## After
image

image


packages/types/src/event.ts
```diff
@@ -29,6 +29,12 @@ export const EventWriteSchema = EventSchema.partial({
 
 export type EventWrite = z.infer<typeof EventWriteSchema>
 
+export const EventFull = EventSchema.extend({
+  eventCommittees: z.array(EventCommitteeSchema),
+})
+
+export type EventFull = z.infer<typeof EventFull>
```
henrikskog, line 36: This is an important change. I'm not sure if it's the best approach, but it makes some sense. We need the EventSchema to match the database schema because it is used by the event repository. However, the UI needs a type that includes other relations. Feedback here would be nice

junlarsen: Let's name it EventWithCommittees

apps/dashboard/src/app/(dashboard)/event/edit-form.tsx
```diff
@@ -63,10 +75,19 @@ export const useEventEditForm = ({ onSubmit, label = "Opprett arrangement", defa
...
         data: committees.map((committee) => ({ value: committee.id, label: committee.name })),
+        onChange: (data) => {
+          console.log("test", data)
+        },
+        onSubmit: (data) => {
+          console.log("test", data)
+        },
+        onClick: (data) => {
+          console.log("test", data)
+        },
```
junlarsen, line 90: remove

apps/dashboard/src/app/form.tsx
```diff
@@ -35,6 +37,29 @@ type InputFieldContext<T extends FieldValues> = {
...
+      <Controller
+        control={control}
+        name={name}
+        render={({ field }) => (
+          <MultiSelect
+            {...props}
+            error={state.errors[name] && <ErrorMessage errors={state.errors} name={name} />}
+            onChange={field.onChange}
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore
```
junlarsen, line 54: Always use ts-expect-error instead of tsignore
```suggestion
            // @ts-expect-error
```

apps/dashboard/src/components/molecules/company-name/event-committees.tsx
```diff
@@ -0,0 +1,25 @@
+import { Committee, EventCommittee } from "@dotkomonline/types"
+import { Anchor } from "@mantine/core"
+import { FC } from "react"
+
+interface Props {
+  committees: Committee[]
+}
+export const EventCommittees: FC<Props> = ({ committees }) => {
+  if (!committees.length) {
+    return "Ingen arrangører"
```
junlarsen, line 10: Wrap in <Text> maybe?

packages/core/src/modules/event/event-service.ts
```diff
@@ -3,11 +3,12 @@ import { Attendance, AttendanceWrite, Event, EventWrite } from "@dotkomonline/ty
 import { NotFoundError } from "../../errors/errors"
 import { Cursor } from "../../utils/db-utils"
 import { AttendanceRepository } from "./attendance-repository"
+import { EventCommitteeRepository } from "./event-committee-repository"
 import { EventRepository } from "./event-repository"
 
 export interface EventService {
-  createEvent(eventCreate: EventWrite): Promise<Event>
-  updateEvent(id: Event["id"], payload: Omit<EventWrite, "id">): Promise<Event>
+  createEvent(eventCreate: EventWrite, eventCommittees: string[]): Promise<Event>
```
junlarsen, line 10: Shouldn't include edges on create in my opinion, better to do the setCommittees stuff in the caller.

packages/core/src/modules/event/event-service.ts
```diff
@@ -20,19 +21,25 @@ export interface EventService {
...
     private readonly eventRepository: EventRepository,
-    private readonly attendanceRepository: AttendanceRepository
+    private readonly attendanceRepository: AttendanceRepository,
+    private readonly committeeOrganizerRepository: EventCommitteeRepository
   ) {}
 
-  async createEvent(eventCreate: EventWrite): Promise<Event> {
+  async createEvent(eventCreate: EventWrite, committees: string[]): Promise<Event> {
     const event = await this.eventRepository.create(eventCreate)
+
```
junlarsen, line 30: A lot of unnecessary whitespace added here maybe you forgot to swap eslint configs

packages/db/src/db.generated.d.ts
```diff
@@ -186,8 +182,8 @@ export interface DB {
   attendee: Attendee;
   committee: Committee;
   company: Company;
-  "drizzle.DrizzleMigrations": DrizzleDrizzleMigrations;
```
junlarsen, line 189: interesting, why was this table even here in the first place I wonder (we dont use drizzle)

packages/gateway-trpc/src/modules/event/event-router.ts
```diff
@@ -1,23 +1,31 @@
...
-  }),
+  create: protectedProcedure
+    .input(
+      z.object({
+        event: EventWriteSchema,
+        committees: z.array(EventCommitteeSchema.shape.committeeId),
+      })
+    )
+    .mutation(({ input, ctx }) => {
+      return ctx.eventService.createEvent(input.event, input.committees)
```
junlarsen, line 17: Do the setCommittes call here instead

packages/types/src/committee.ts
```diff
@@ -2,7 +2,7 @@ import { z } from "zod"
 
 export const CommitteeSchema = z.object({
   id: z.string().ulid(),
-  createdAt: z.date(),
+  createdAt: z.coerce.date(),
```
junlarsen, line 5: Why is this needed? we use superjson so dates shouldn't have to be coerced into string anywhere

packages/core/src/modules/event/event-committee-repository.ts
```diff
@@ -0,0 +1,48 @@
...
+  async setCommittees(eventId: Event["id"], committees: Committee["id"][]): Promise<void> {
+    // remove all committees for event
+    await this.db.deleteFrom("eventCommittee").where("eventId", "=", eventId).execute()
+
+    if (committees.length === 0) return
+    // add all committees for event
+    const committeesToAdd = committees.map((committeeId) => ({ eventId, committeeId }))
+
+    await this.db.insertInto("eventCommittee").values(committeesToAdd).execute()
+  }
```
junlarsen, line 47: prefer naming methods that connect edges (like connecting committee to event) add instead of set.

I think this is better represented as two different functions named add/remove, and determining the diff should be done in the service

packages/core/src/modules/event/event-committee-service.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export class EventCommitteeServiceImpl implements EventCommitteeService {
+  constructor(private readonly committeeOrganizerRepository: EventCommitteeRepositoryImpl) {}
+
+  async getCommitteesForEvent(eventId: Event["id"]): Promise<Committee[]> {
+    const committees = await this.committeeOrganizerRepository.getAllCommittees(eventId)
+    return committees
+  }
+
+  async setEventCommittees(eventId: Event["id"], organizers: Committee["id"][]): Promise<void> {
```
junlarsen, line 16: This can stay as "set" but we should do the difference logic here as explained in the repository.

packages/core/src/modules/event/event-company-repository.ts
```diff
@@ -64,6 +63,6 @@ export class EventCompanyRepositoryImpl implements EventCompanyRepository {
       query = query.orderBy("id", "desc")
     }
     const events = await query.execute()
-    return events.map(mapToEvent)
+    return events.map((e) => EventSchema.parse(e))
```
junlarsen, line 66: Can we not do mapToEvent here? alternatively mapToEventWithCommittees?

packages/core/src/modules/event/event-repository.ts
```diff
@@ -27,30 +57,78 @@ export class EventRepositoryImpl implements EventRepository {
       .where("id", "=", id)
       .returningAll()
       .executeTakeFirstOrThrow()
-    return mapToEvent(event)
+    return EventSchema.parse(event)
```
junlarsen, line 60: create mapToEventWithCommittees, there should at least be a mapper function available so we don't write .parse() everywhere for consistency

junlarsen (changes requested):

packages/types/src/committee.ts
henrikskog: Did not mean for the latest commit to be reviewed, was mostly to show code to Anhkha (: so this was from testing earlier

henrikskog: but yeah my bad, sorry

packages/core/src/modules/event/event-committee-repository.ts
henrikskog: Good feedback!

packages/core/src/modules/event/event-company-repository.ts
henrikskog: Fixed

packages/db/src/db.generated.d.ts
henrikskog: No idea! Weird

packages/types/src/committee.ts
junlarsen: ah no problem

junlarsen (approved):
LGTM great work!

PR #610, opened 2023-10-14
Move pagination and ordering logic into helper function

junlarsen:
Also renamed it to orderedQuery to make more sense out of its name.

Maybe we should add sort/filter parameters to it too?


henrikskog (approved):
lgtm

PR #621, opened 2023-10-18
Make use of id types consistent

junlarsen:
Less visual noise without the ["id"] and makes it all consistent


junlarsen:
ignore the git log I branched off an already merged pr

PR #626, opened 2023-10-18
Sync calendar with events

joleeee:
# Screenshot 📸
image

# No auth endpoints
To test it locally you go to
- `http://localhost:3000/api/cal/all`
- ~~`http://localhost:3000/api/cal/user/01HB64XF7WXBPGVQKFKFGJBH4D` (or another id ofc)~~
- `http://localhost:3000/api/cal/event/01HB64TWZK1N8ABMH8JAE12101`

# Auth (new :))))))
To get access to a users calendar, that signed in user must go to `/api/cal/sign` and then use the returned token in `/api/cal/user/${token}`

```sh
# not authed
$ curl http://localhost:3000/api/cal/sign
{"message":"Not signed in"}

# getting the token
$ curl 'http://localhost:3000/api/cal/sign' -H 'Cookie: next-auth.session-token=...'
{"token":"eyJhbGciOiJIUzI1NiJ9.MDFIUTZSNkdFWVcyWjNWV1ZCU0FYWFRLQlo.XklF4gPVajozOE8ZsImhHFZXzZc-fF6qQgeX5xNvOoM"}

# the contents
$ base64 -d <<< $(echo MDFIUTZSNkdFWVcyWjNWV1ZCU0FYWFRLQlo)
01HQ6R6GEYW2Z3VWVBSAXXTK

# downloading the calendar (can add this link to calendar app, google calendar, etc)
$ curl http://localhost:3000/api/cal/user/eyJhbGciOiJIUzI1NiJ9.MDFIUTZSNkdFWVcyWjNWV1ZCU0FYWFRLQlo.XklF4gPVajozOE8ZsImhHFZXzZc-fF6qQgeX5xNvOoM
BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//sebbo.net//ical-generator//EN
NAME:01HQ6R6GEYW2Z3VWVBSAXXTKBZ online kalender
<...>

```


kharann:
perhaps remove the todo comments and make linear issues of them when you merge inn this PR?

packages/core/src/modules/event/event-repository.ts
```diff
@@ -38,6 +39,26 @@ export class EventRepositoryImpl implements EventRepository {
...
+      .leftJoin("attendee", "attendee.attendanceId", "attendance.id")
+      .select("attendance.eventId")
+      .where("attendee.userId", "=", userId)
+      .groupBy("attendance.eventId")
+      .execute()
+
+    const eventPromises = event_ids
+      .map(({ eventId }) => eventId)
+      .filter((id): id is string => id !== null)
+      .map(async (id) => this.getById(id))
```
junlarsen, line 55: Is it possible to change this to a left join onto events? I think the current code causes an 1+N query

packages/gateway-edge-nextjs/src/cal/cal.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  events.forEach(event => {
+    instance.createEvent(toICal(event));
+    instance.createEvent(toICal(toRegistration(event)));
+  });
+
+  res.status(200).send(instance.toString())
+}
+
+function toICal(event: { start: Date, end: Date, title: string, description: string | null, location: string | null, id: string }): ICalEventData {
```
junlarsen, line 88: You can refactor this type to `Pick<Event, ...>` to keep a single source of truth for the Event attributes

packages/gateway-edge-nextjs/src/cal/cal.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+}
+
+function toRegistration(event: { start: Date, end: Date, title: string, description: string | null, location: string | null, id: string }): { start: Date, end: Date, title: string, description: string | null, location: string | null, id: string } {
+  // 5 days before
+  // TODO when db has this, we can use the actual start value, this is just for testing
+  const start = new Date(event.start.getTime() - 5 * 24 * 60 * 60 * 1000);
+
+  const title = `Påmelding for ${event.title}`;
+  const location = "På OW";
```
junlarsen, line 105: In the future it would be sick we could make this point to the URL of the event on the website

junlarsen (commented):
Looks good, just a few small refactor comments!

packages/gateway-edge-nextjs/src/cal/cal.ts
joleeee: Den gjør i prinsipp det allerede :)) 
image

packages/gateway-edge-nextjs/src/cal/cal.ts
joleeee: Ok i did it, pls review

packages/gateway-edge-nextjs/src/cal/cal.ts
```diff
@@ -0,0 +1,117 @@
...
+  router: appRouter,
+  ctx: await createContextInner({
+    auth: null,
+  }),
+  transformer, // optional - adds superjson serialization
+})
+
+function eventUrl(event: Pick<Event, "id">) {
+  // a better to to get/configure the url?
+  return `https://new.online.ntnu.no/events/${event.id}`
```
junlarsen, line 17: New url is currently dev.web.online.ntnu.no but it doesn't really matter

packages/gateway-edge-nextjs/src/cal/cal.ts
```diff
@@ -0,0 +1,117 @@
...
+    start: event.start,
+    end: event.end,
+    summary: event.title,
+    description: event.description,
+    location: event.location,
+    url: eventUrl(event),
+  }
+}
+
+// function toRegistration(event: { start: Date, end: Date, title: string, description: string | null, location: string | null, id: string }): { start: Date, end: Date, title: string, description: string | null, location: string | null, id: string } {
```
junlarsen, line 97: Can remove this line

package.json
```diff
@@ -40,14 +40,14 @@
...
+    "prettier": "^3.2.4",
+    "prettier-plugin-tailwindcss": "^0.5.11",
+    "ts-node": "^10.9.2",
     "tsc-watch": "^6.0.4",
-    "turbo": "^1.10.16",
-    "turbo-ignore": "^1.10.16",
-    "typescript": "^5.2.2",
+    "turbo": "^1.11.3",
+    "turbo-ignore": "^1.11.3",
+    "typescript": "^5.3.3",
```
junlarsen, line 50: Did the dependencies get bumped automatically or something here?

junlarsen (changes requested):

package.json
joleeee: eghrhrhm someone updated trpc in some "modules" (or what they are called) but not in this one so i had to update it, but i can udno everything other than trpc

packages/core/src/modules/event/event-repository.ts
joleeee: it should now use 2 queries

joleeee:
yolo merge ?

packages/core/src/modules/event/event-repository.ts
```diff
@@ -40,6 +41,22 @@ export class EventRepositoryImpl implements EventRepository {
     return events.map((e) => mapToEvent(e))
   }
 
+  async getAllByUserAttending(userId: string): Promise<Event[]> {
+    const event_ids_query = await this.db
```
henrikhorluck, line 45: This should just do one query, not fetch ids, and then fetch Everything related to those id's

packages/core/src/modules/payment/payment-service.ts
```diff
@@ -199,8 +199,9 @@ export class PaymentServiceImpl implements PaymentService {
   }
 
   async refundPayment(payment: Payment, product: Product): Promise<void> {
-    const paymentProvider = product.paymentProviders.find((p) => p.paymentProviderId === payment.paymentProviderId)
-      ?.paymentProvider
+    const paymentProvider = product.paymentProviders.find(
```
henrikhorluck, line 202: This seems unrelated. If it was from a manual change please avoid doing so in the future, otherwise it might be from a dependency-update

packages/gateway-edge-nextjs/src/cal/cal.ts
```diff
@@ -0,0 +1,150 @@
...
+  router: appRouter,
+  ctx: await createContextInner({
+    auth: null,
+  }),
+  transformer, // optional - adds superjson serialization
+})
+
+function eventUrl(event: Pick<Event, "id">) {
+  // a better to to get/configure the url?
+  return `https://dev.web.online.ntnu.no/events/${event.id}`
```
henrikhorluck, line 20: This should be fixed before merge. @junlarsen what is the next-equivalent of Django's `reverse`

packages/gateway-edge-nextjs/src/cal/cal.ts
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// ALL events
+export async function CalendarAll(req: NextApiRequest, res: NextApiResponse) {
+  if (req.method !== "GET") {
```
henrikhorluck, line 25: Are methods not handled at the routing-level? You should not need to do this kind of error-handling yourself IMO

packages/core/src/modules/event/event-repository.ts
```diff
@@ -54,8 +71,17 @@ export class EventRepositoryImpl implements EventRepository {
     const events = await query.execute()
     return events.map((e) => mapToEvent(e))
   }
+
   async getById(id: string): Promise<Event | undefined> {
     const event = await this.db.selectFrom("event").where("id", "=", id).selectAll().executeTakeFirst()
     return event === undefined ? undefined : mapToEvent(event)
   }
+
+  async getByIds(id: string[]): Promise<Event[]> {
```
henrikhorluck, line 80: We should not need this.

henrikhorluck (changes requested):

packages/core/src/modules/payment/payment-service.ts
joleeee: undone

packages/gateway-edge-nextjs/src/cal/cal.ts
joleeee: i updated it, please re-review

packages/core/src/modules/event/event-repository.ts
joleeee: implemented

packages/core/src/modules/event/event-repository.ts
```diff
@@ -40,6 +41,18 @@ export class EventRepositoryImpl implements EventRepository {
...
   }
 
+  async getAllByUserAttending(userId: string): Promise<Event[]> {
+    const eventsResult = await this.db
+      .selectFrom("attendance")
+      .leftJoin("attendee", "attendee.attendanceId", "attendance.id")
+      .innerJoin("event", "event.id", "attendance.eventId")
+      .selectAll("event")
+      .where("attendee.userId", "=", userId)
+      .groupBy("event.id")
```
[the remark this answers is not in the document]
joleeee: implemented

packages/gateway-edge-nextjs/src/cal/cal.ts
joleeee: cant figure out how to do it in nextjs, maybe @junlarsen knows

joleeee:
Don't understand why i get lint errors in `packages/db/src/db.generated.d.ts`, I didn't change it??

packages/gateway-edge-nextjs/package.json
```diff
@@ -16,6 +16,13 @@
   "dependencies": {
     "@dotkomonline/core": "workspace:*",
     "@dotkomonline/db": "workspace:*",
+    "@dotkomonline/gateway-trpc": "workspace:^",
+    "@dotkomonline/types": "workspace:^",
+    "@trpc/react-query": "^10.45.0",
+    "ical-generator": "^5.0.1",
+    "jsonwebtoken": "^9.0.2",
+    "@types/jsonwebtoken": "^9.0.5",
```
junlarsen, line 24: all `@types/*` dependencies should be moved to devDependencies

packages/gateway-edge-nextjs/src/cal/cal.ts
junlarsen: If we switch to using these we should be able to

https://nextjs.org/docs/app/building-your-application/routing/route-handlers but I think this is fine for now.

junlarsen (commented):
So if I understand properly, the flow is:

1. User visits /api/cal/sign to get a token
2. The user then points the calendar app to /api/cal/user/{id}?user=JWT
3. When the calendar attempts to fetch the data, the /api/cal/user/{id} endpoint does the following
    - Get the JWT from the query string
    - Validate the JWT
    - If it's valid, give the calendar data

junlarsen:
> Don't understand why i get lint errors in `packages/db/src/db.generated.d.ts`, I didn't change it??

I don't know, maybe you can try to `pnpm lint -- --fix` from root?

packages/gateway-edge-nextjs/src/cal/cal.ts
```diff
@@ -0,0 +1,155 @@
...
+const helpers = createServerSideHelpers({
+  router: appRouter,
+  ctx: await createContextInner({
+    auth: null,
+  }),
+  transformer, // optional - adds superjson serialization
+})
+
+function eventUrl(req: NextApiRequest, event: Pick<Event, "id">) {
+  const proto = req.headers["x-forwarded-proto"] || "http"
```
henrikhorluck, line 19:
```suggestion
  const proto = req.headers["x-forwarded-proto"] || "https"
```

I find it unlikely that this will be tested locally that often—I'd rather have some kind of tests for this.

packages/gateway-edge-nextjs/src/cal/cal.ts
```diff
@@ -0,0 +1,155 @@
...
+  const host = req.headers["x-forwarded-host"] || "online.ntnu.no"
```
henrikhorluck, line 20: okay I get that this _works_ in prod/dev/whatever, but this wouldn't work locally? I am fine with merging this without sorting it out, but we should add a config or something for `BASE_URL`/ `HOST`

packages/gateway-edge-nextjs/src/cal/cal.ts
henrikhorluck: So you just export as `GET` instead of a default export?

henrikhorluck (commented):
Would love if you wrote some tests as well :D

packages/gateway-edge-nextjs/src/cal/cal.ts
joleeee: I think it should always work because of nextjs magic, read bottom here (doesn't say it *always* works, but doesn't have a disclaimer either) https://nextjs.org/docs/app/api-reference/functions/headers#ip-address

packages/gateway-edge-nextjs/src/cal/cal.ts
joleeee: testing for what?

henrikhorluck: That it returns an usable URI?

henrikhorluck: Nvm, I didn’t think clear enough about how this is deployed

packages/gateway-edge-nextjs/src/cal/cal.ts
joleeee: it is set locally for me 🙃

joleeee: i dont see why it would change behavior in prod, everything else being equal?

junlarsen:
If you want to try testing the service methods, you can look at some of the tests for job-listing. I think it's pretty similar to what you want

https://github.com/dotkom/monoweb/blob/main/packages/core/src/modules/job-listing/__test__/job-listing.e2e-spec.ts

packages/gateway-edge-nextjs/src/cal/cal.ts
joleeee: image

Given we already use this in other places, I think that should be an issue of its own

henrikhorluck (approved):

joleeee:
yolo

PR #648, opened 2023-10-23
Set up integration testing with Testcontainers

junlarsen:
## What's new?

This patch has integration testing with Testcontainers which means that we can now run tests against a real PostGreSQL database.

- It uses the PostGreSQL image from #707 
- It spins up a new database for each test (it's slow, but integration tests do not need to be fast)
- It runs with `pnpm test:it`

## Changes

- All database migrations and files imported in migrations are now .js files in order to make Vitest able to process them at runtime. This doesn't really have a big impact because the `db` parameter was `Kysely<any>` anyways.
- Fixed a bug with UserService.getById/Subject throwing instead of returning undefined
- Fixed a buf where UserService edit was not set to accept Partial


kharann:
This is cool :D

packages/core/vitest-integration.setup.ts
```diff
@@ -0,0 +1,25 @@
...
+    .start()
+  process.env.DATABASE_URL = container.getConnectionUri()
+  const env = createEnvironment()
+  const kysely = createKysely(env)
+  const migrator = createMigrator(kysely, new URL("node_modules/@dotkomonline/db/src/migrations", import.meta.url))
+  await migrator.migrateToLatest()
+  containers.push(container)
+})
+
+process.on("exit", async () => {
```
henrikskog, line 23: If this code is expected to run, I believe the event that should be listened to is "beforeExit". No additional work will be done after the "exit" event, so the container.stop calls will not run.

See https://nodejs.org/api/process.html#event-exit and https://nodejs.org/api/process.html#event-beforeexit

apps/db-migrator/src/db.ts
```diff
@@ -1,10 +1,10 @@
-import { CockroachDialect, type Database } from "@dotkomonline/db"
-import { Kysely, CamelCasePlugin } from "kysely"
+import { type Database } from "@dotkomonline/db"
+import { Kysely, CamelCasePlugin, PostgresDialect } from "kysely"
 import pg from "pg"
 import { env } from "@dotkomonline/env"
 
 export const db = new Kysely<Database>({
-  dialect: new CockroachDialect({
+  dialect: new PostgresDialect({
```
henrikskog, line 7: Did we use CockroachDB before?

kharann: yes we did since it was free 🤑

packages/db/src/index.ts
henrikskog, line 1: I'm a bit puzzled by the logic here with the kysely global attribute not in dev etc. But looks good afaik

packages/core/vitest-integration.setup.ts
junlarsen: Ahh good catch

apps/db-migrator/src/db.ts
```diff
@@ -1,10 +1,10 @@
-import { CockroachDialect, type Database } from "@dotkomonline/db"
-import { Kysely, CamelCasePlugin } from "kysely"
+import { type Database } from "@dotkomonline/db"
+import { Kysely, CamelCasePlugin, PostgresDialect } from "kysely"
 import pg from "pg"
 import { env } from "@dotkomonline/env"
 
 export const db = new Kysely<Database>({
```
henrikskog, line 6: why not use `createKysely` from `packages/db/src/index.ts` here?

henrikskog (changes requested):
This is sick! Left a few comments 👍

henrikskog (changes requested):
This is sick! Left a few comments 👍

apps/db-migrator/src/db.ts
henrikskog: aha!

packages/auth/src/auth-options.ts
```diff
@@ -50,6 +50,9 @@ export const getAuthOptions = ({
     async session({ session, token }) {
       if (token.sub) {
         const user = await core.userService.getUserBySubject(token.sub)
+        if (user === undefined) {
+          throw new Error(`Found no matching user for ${token.sub}`)
```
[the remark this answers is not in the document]
junlarsen: yep!

junlarsen:
> This is sick! Left a few comments 👍

Good insights, I'll revisit this later today!

packages/core/vitest-integration.setup.ts
```diff
@@ -0,0 +1,25 @@
+import { beforeEach } from "vitest"
+import { PostgreSqlContainer, type StartedPostgreSqlContainer } from "@testcontainers/postgresql"
+import { createMigrator, createKysely } from "@dotkomonline/db"
+import { createEnvironment } from "@dotkomonline/env"
+
+const containers: StartedPostgreSqlContainer[] = []
+
+beforeEach(async () => {
```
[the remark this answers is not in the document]
junlarsen: Think we can consider that in the future if we hit any roadbumps

henrikskog (approved):
Super nice, awesome to get this in

PR #659, opened 2023-10-26
Add and remove event attendees

NjaalSoerland:
Some styling that needs to be done, but it makes more sense to fix that when dealing with attendee pools. So this should be merged first


apps/gateway-email/package.json
NjaalSoerland, line 1: necessary to be able to run pnpm dev. Didn't mean to push but don't know how to make it work locally with this script

apps/gateway-email/package.json
```diff
@@ -4,7 +4,6 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "dev": "docker run -p 9000:8080 -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -e AWS_REGION=eu-north-1 gateway-email:latest",
```
NjaalSoerland, line 7: necessary to be able to run pnpm dev. Didn't mean to push but don't know how to make it work locally with this script

junlarsen: Can you rename this and dispatcher-auth's dev scripts to "docker:dev"?

junlarsen (changes requested):
LGTM, just change the script names and rebase on main first

apps/dashboard/src/components/GenericSearch.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { Autocomplete } from "@mantine/core"
+
+interface GenericSearchProps<T> {
+  onSearch: (query: string) => void
+  onSubmit: (item: T) => void
+  items: T[]
+  dataMapper: (item: T) => string
+}
+
+const GenericSearch = <T,>({ onSearch, onSubmit, items, dataMapper }: GenericSearchProps<T>) => {
```
junlarsen, line 11: What if we make GenericSearchProps also include the mantine Autocomplete props so you can override placeholder if you want to?

junlarsen (approved):
LGTM with a suggestion maybe for anoter PR

PR #663, opened 2023-10-27
Move report interest form into Monoweb

junlarsen:
Also uses our email handler

image
image
image


apps/rif/README.md
henrikskog, line 1: Looks like this is left over from dashboard ;)

henrikskog (changes requested):
Awesome! When testing out the form I noticed that the errors are in english. Could we do those in norwegian as well?

henrikskog (approved):
Great job!!

PR #680, opened 2023-10-29
Add event extras system with management in dashboard

henrikskog:
Includes core logic for adding choices for attendees, but no interface for that as its not something we need on the dashboard (AFAIK)

CLOSES DOT-247
CLOSES DOT-250

image
image
image


kharann:
I do not understand, can an event have more then 1 extra? Also I think that:

When adding extras you could add a template through a select with some standard extras such as pizza/sushi etc.

These values could be hard coded in the frontend (the template/common values).

kharann:
Yeh yeh I agree kjør på

henrikskog:
Aight this should be ready now

can be tested out locally with DATABASE_URL="postgres://web:<passord>@ep-mute-breeze-50080782.eu-central-1.aws.neon.tech/extras?sslmode=require" # feature branch 1

junlarsen (changes requested):
Code looks good! I'm wondering if we maybe should use icon-only buttons for edit and delete? 

From a user experience perspective; I think the yellow and red buttons attract too much attention when they are likely not the action the user is going to perform.

I suggest we change them to use Mantine ActionIcon buttons

https://mantine.dev/core/action-icon/

henrikskog:
image

Obviously the design is not very appealing, it has not been the focus. Not sure what I think about this either. What do you think?

junlarsen (approved):

PR #688, opened 2023-10-31
Add stub for pg-native for dashboard

junlarsen:
Next.js tries to include pg-native which is a module that uses C++ addons into the webpack build on the server (because we have transpilePackages)

It errors because it thinks the package will end up on the client, but it won't so for webpack we will just stub it


henrikskog (approved):
Very nice!

PR #689, opened 2023-10-31
wip webshop

henrikskog:
image
image
image

There is also added a webhook that adds orders to the dashboard on the checkout.session.completed event such that prokom can mark the orders as delivered or not.


kharann:
bit stale any updates?

PR #695, opened 2023-11-01
Add missing public environment variables

junlarsen:


jotjern (approved):
good one mats keep it up

PR #699, opened 2023-11-01
Add study years to users

NjaalSoerland:
Adds study years as an enum from -1 to 6 associated with all types of membership a user can have.
Adds ability to edit study years for users through the dashboard


packages/types/src/user.ts
```diff
@@ -15,3 +16,31 @@ export const UserWriteSchema = UserSchema.omit({
...
+interface StudyYears {
+  [-1]: string
+  0: string
+  1: string
+  2: string
+  3: string
+  4: string
+  5: string
+  6: string
+  [key: number]: string
```
junlarsen, line 29: Do we need other keys as well? I think we can remove the `[key: number]: string`

junlarsen (commented):
I have one small question, but otherwise LGTM

packages/types/src/user.ts
NjaalSoerland: it works without it, but i get the following warning in `apps/dashboard/src/app/(dashboard)/user/page.tsx` when it's not included.
"Element implicitly has an 'any' type because expression of type 'number' can't be used to index type 'StudyYears'."
Using any other numbers won't work anyways, so `[key: number]: string` is just there to avoid it being interpreted as 'any'

henrikskog (approved):
Niiice!

PR #705, opened 2023-11-06
Create backend and dashboard for creating articles

junlarsen:
Does currently not handle tags


packages/types/src/article.ts
```diff
@@ -0,0 +1,36 @@
...
+export const ArticleTagWrite = ArticleTagSchema
+export type ArticleTagWrite = z.infer<typeof ArticleTagWrite>
+
+export const ArticleSchema = z.object({
+  id: z.string().ulid(),
+  createdAt: z.date(),
+  updatedAt: z.date(),
+  title: z.string().min(1),
+  author: z.string().min(1),
+  photographer: z.string().min(1),
```
junlarsen, line 19: Could *technically* have a relation onto OwUser here, but I don't think todays use of this field really warrants it. It also makes the query for fetching the names more complex as we have to go through AWS Cognito

packages/db/src/migrations/0020_add_article.ts
```diff
@@ -0,0 +1,29 @@
+import { type Kysely, sql } from "kysely"
+import { createTableWithDefaults } from "../utils"
+
+export async function up(db: Kysely<any>) {
+  await createTableWithDefaults("articles", { id: true, createdAt: true, updatedAt: true }, db.schema)
+    .addColumn("title", "text", (col) => col.notNull())
+    .addColumn("author", "text", (col) => col.notNull())
+    .addColumn("photographer", "text", (col) => col.notNull())
+    .addColumn("image_url", "text", (col) => col.notNull())
```
henrikskog, line 9: Should it be mandatory to upload an image for the article?

packages/db/src/migrations/0020_add_article.ts
```diff
@@ -0,0 +1,29 @@
+import { type Kysely, sql } from "kysely"
+import { createTableWithDefaults } from "../utils"
+
+export async function up(db: Kysely<any>) {
+  await createTableWithDefaults("articles", { id: true, createdAt: true, updatedAt: true }, db.schema)
+    .addColumn("title", "text", (col) => col.notNull())
+    .addColumn("author", "text", (col) => col.notNull())
+    .addColumn("photographer", "text", (col) => col.notNull())
```
henrikskog, line 8: Won't there be situations where there is no obvious photographer for an image? Such as ai generated images etc. Do we want to enforce having a photographer on the db level?

packages/db/src/migrations/0020_add_article.ts
```diff
@@ -0,0 +1,29 @@
...
+    .addColumn("content", "text", (col) => col.notNull())
+    .execute()
+  await db.schema
+    .createTable("article_tags")
+    .addColumn("name", "text", (col) => col.primaryKey())
+    .execute()
+  await db.schema
+    .createTable("article_tag_link")
+    .addColumn("article", sql`ulid`, (col) => col.references("articles.id").notNull())
+    .addColumn("tag", "text", (col) => col.references("article_tags.name").notNull())
```
henrikskog, line 21: cascade on delete?

packages/types/src/article.ts
henrikskog: In my opinion it's kind of unfortunate to not establish a relation to the given user for the author field. In the future we may want to do something like abakus does with profiles. At that point it makes sense to link to the profile of the author in the article. Like this: https://abakus.no/articles/488-weekly-uke-44. 

At least I would like a `select` for the author field in the write form. This way we could link them up in the future.

packages/core/src/modules/article/article-tag-link-repository.ts
```diff
@@ -0,0 +1,27 @@
+import { type ArticleId, type ArticleTagName } from "@dotkomonline/types"
+import { type Kysely } from "kysely"
+import { type Database } from "@dotkomonline/db"
+
+export interface ArticleTagLinkRepository {
```
henrikskog, line 5: should these not return the resource that was added/removed?

apps/dashboard/src/modules/article/mutations/use-create-article-mutation.ts
```diff
@@ -0,0 +1,31 @@
...
+    onMutate: () => {
+      notification.loading({
+        title: "Lager artikkel...",
+        message: "Artikkelen blir opprettet, og du vil bli videresendt til artikkelsiden.",
+      })
+    },
+    onSuccess: (data) => {
+      notification.complete({
+        title: "Artikkel opprettet",
+        message: `Artikkeln "${data.title}" har blitt opprettet.`,
```
henrikskog, line 19: Artikkeln

apps/dashboard/src/app/(dashboard)/article/write-form.tsx
```diff
@@ -0,0 +1,60 @@
...
+    label,
+    fields: {
+      title: createTextInput({
+        label: "Tittel",
+        placeholder: "Fadderuka 2023",
+        withAsterisk: true,
+      }),
+      author: createTextInput({
+        label: "Forfattere",
+        placeholder: "Ola Nordmann, Trond-Viggo Torgersen",
```
henrikskog, line 30: I would like a select component for this

henrikskog, line 30: And then, when we have the user, why not have the ow_user id on the article as well?

packages/db/src/migrations/0020_add_article.ts
```diff
@@ -0,0 +1,29 @@
+import { type Kysely, sql } from "kysely"
+import { createTableWithDefaults } from "../utils"
+
+export async function up(db: Kysely<any>) {
+  await createTableWithDefaults("articles", { id: true, createdAt: true, updatedAt: true }, db.schema)
+    .addColumn("title", "text", (col) => col.notNull())
+    .addColumn("author", "text", (col) => col.notNull())
```
henrikskog, line 7: author_ow_user_id?

henrikskog (approved):
Looks nice! Left some comments

henrikskog (changes requested):
Didn't mean to approve on the last review ;)

packages/db/src/migrations/0020_add_article.ts
junlarsen: I think it's reasonable to say so, I don't think there has been an article on OW without an image for as long as I've been a student

packages/db/src/migrations/0020_add_article.ts
junlarsen: I think in these cases it should be attributed to the image generator, what do you think?

junlarsen: I think it's probably nice to have a required image alt text field here too

packages/db/src/migrations/0020_add_article.ts
junlarsen: Good catch, agree

apps/dashboard/src/app/(dashboard)/article/write-form.tsx
junlarsen: Hard to do with the current state of ow_user would have to search for their ulids right now.

packages/db/src/migrations/0020_add_article.ts
henrikskog: Fair enough!

packages/db/src/migrations/0020_add_article.ts
henrikskog: Yeah it actually seems nice to always have a photographer and alt text

packages/core/src/modules/article/article-tag-link-repository.ts
henrikskog: Ah yeah true. Okay to leave it out here.

apps/dashboard/src/app/(dashboard)/article/write-form.tsx
henrikskog: Should we sync up the attributes of users in cognito with ow_user maybe? It seems very unfortunate to not be able to do stuff like this. However, I agree that it does not need to be a part of this issue, so text input is fine!

henrikskog: I had a bit of a misunderstanding about what was possible to do with ow_user. You can ignore the comment

packages/db/src/migrations/0020_add_article.ts
henrikskog: We don't need this at this moment, you can ignore the comment.

henrikskog (approved):
You can look over the comments, but I think this is ready to be merged for the purpose of a MVP.

junlarsen:
Oh I completely forgot about this pr, I'll have a closer look later today

PR #707, opened 2023-11-06
Add public ecr repository for pgx-ulid

junlarsen:
Adds a postgres image with the pgx_ulid extension installed. It's pushed to public ECR


jotjern (approved):
aaaaaaaaaaamazing

PR #708, opened 2023-11-06
Move migrator into tools

junlarsen:
I think adding a tools folder to the top level makes sense. This will be things that are meant to be ran locally and not deployed anywhere (unlike apps)

Didn't touch sanity since it's going bye bye soon

Interested in hearing what you think @henrikskog @NjaalSoerland @kharann


kharann:
I think its fair. The t3 turbo example does it https://github.com/t3-oss/create-t3-turbo/tree/main/tooling

henrikskog:
Sounds good! I think that makes it more clear 👍

henrikskog (approved):

PR #711, opened 2023-11-06
Create paginated collection builder

junlarsen:
Problems this solves outlined in https://linear.app/dotkom/issue/DOT-426/make-all-backend-endpoints-return-a-collection-type

All the client queries now have a cursor argument they can use to determine if there is more data coming

The end goal is to replace every orderedQuery with paginatedQuery


henrikskog:
This is super cool. Would be nice to have a test here maybe?

henrikskog (approved):

PR #712, opened 2023-11-08
Fix career page navbar bug

madsab:
Closes: #...

## Checklist

- [x] I fixed the bug  with the navbar in career page were hovering over "Om oss" or "For bedrifter" would appear behind CareerView component and not be possible to click

## Changelog

- Removed z-index property for div in CareerView

## How to test
- Run monoweb
- Navigate to <code>localhost:3000/career</code>
- Hover over "Om oss" and "For bedrifter"


henrikskog (approved):
Nice fix, this was super annoying

PR #719, opened 2023-11-10
Add missing import for NotFoundError

junlarsen:
Seems like intellij autoimport failed me :sob:


henrikskog (approved):
I wish I caught that haha

PR #720, opened 2023-11-10
Remove job listing from interest form

junlarsen:
Refer to response email from bedkom with feedback for the page.

- Removed job listing from the schema, with a text message instructing companies to directly contact bedkom
- Changed the input form color to white


madsab (approved):
LGTM!

PR #774, opened 2023-11-27
allow dashboard origin in trpc requests

henrikskog:
Dashboard running on dev.dashboard.online.ntnu.no should be able to make requests to web


apps/web/src/middlewares/cors.ts
```diff
@@ -3,7 +3,7 @@ import Cors from "cors"
 import { type NextApiHandler, type NextApiRequest, type NextApiResponse } from "next"
 
 const cors = Cors({
-  origin: ["http://localhost:3002"],
+  origin: ["http://localhost:3002", "http://dev.dashboard.online.ntnu.no", "https://dev.dashboard.online.ntnu.no"],
```
junlarsen, line 6: https
```suggestion
  origin: ["http://localhost:3002", "https://dev.dashboard.online.ntnu.no", "https://dev.dashboard.online.ntnu.no"],
```

junlarsen (changes requested):

apps/web/src/middlewares/cors.ts
henrikskog: I added both just out of lazyness, but i remove the http instance now

junlarsen (approved):

PR #778, opened 2024-01-17
Specify Next Auth JWT secret manually

junlarsen:
Next Auth doesn't pick this up automatically anymore


jotjern (approved):
i did a thorough security review of this and i can with 100 percent certainty say it's probably fine idk tho

PR #786, opened 2024-01-29
Invalidate cache on all mutations to simplify code

henrikskog:
See https://trpc.io/docs/v10/client/react/useUtils#invalidate-full-cache-on-every-mutation

It's annoying to keep track of what resources to invalidate after a successful mutation. This way it is always invalidated.

There might be some situations where this is not wanted, but I can't think of one for the dashboard.

Also, `utils.getContext()` to get query client for invalidating cache is deprecated (https://trpc.io/docs/v10/migrate-from-v9-to-v10#queryclient-is-no-longer-exposed-through-trpc-context).

### DIFF
--- a/apps/dashboard/src/utils/trpc.ts
+++ b/apps/dashboard/src/utils/trpc.ts
@@ -34,7 +34,26 @@ export const trpcConfig: CreateTRPCClientOptions<AppRouter> = {
 }
 
 // React query trpc
-export const trpc = createTRPCReact<AppRouter>()
+export const trpc = createTRPCReact<AppRouter>({
+  overrides: {
+    useMutation: {
+      /**
+       * This function is called whenever a `.useMutation` succeeds
+       **/
+      async onSuccess(opts) {
+        /**
+         * @note that order here matters:
+         * The order here allows route changes in `onSuccess` without
+         * having a flash of content change whilst redirecting.
+         **/
+        // Calls the `onSuccess` defined in the `useQuery()`-options:
+        await opts.originalFn()
+        // Invalidate all queries in the react-query cache:
+        await opts.queryClient.invalidateQueries()
+      },
+    },
+  },
+})
 
 /**
  * Inference helpers for input types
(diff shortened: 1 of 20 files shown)


apps/dashboard/src/utils/trpc.ts
```diff
@@ -34,7 +34,26 @@ export const trpcConfig: CreateTRPCClientOptions<AppRouter> = {
...
+      async onSuccess(opts) {
+        /**
+         * @note that order here matters:
+         * The order here allows route changes in `onSuccess` without
+         * having a flash of content change whilst redirecting.
+         **/
+        // Calls the `onSuccess` defined in the `useQuery()`-options:
+        await opts.originalFn()
+        // Invalidate all queries in the react-query cache:
+        await opts.queryClient.invalidateQueries()
```
junlarsen, line 52: ahh this is a spicy one

It's probably fine for now, but we might have to re consider once more modules of the dashboards are in. It could mean a lot of queries on each change.

junlarsen (approved):

apps/dashboard/src/utils/trpc.ts
henrikskog: [documentation link](https://tanstack.com/query/v4/docs/framework/react/guides/query-invalidation)
> When a query is invalidated with invalidateQueries, two things happen:
> 1. It is marked as stale. This stale state overrides any staleTime configurations being used in useQuery or related hooks
> 2. If the query is currently being rendered via useQuery or related hooks, it will also be refetched in the background

This is my understanding of what this changes. 

**Before using this solution for invalidation**
- When going to a page on the dashboard, that data will be kept in cache until data from that same module is updated, because after that update you make sure to invalidate the related data. 

The disadvantage of this is to always have to make sure to invalidate the right data. If a bug makes its way in and someone forgets to invalidate, it leads to a really bad user experience. 

**After using this solution for invalidation**
- When going to a page on the dashboard that data will be kept in cache until **any** data is updated. I think that the data will not be refetched until you actually mount the component with that data again. If you update some `event` data, the queries in other modules are not ran before you render those modules again. But you loose the nice instant load from cache.

I'm not sure it's a big problem that cache of other modules is lost after an update as in many cases you won't be interested in other domains afterwards anyway. But I'm not sure, and in some cases this might not be ideal 

**Another thing**
We should probably remove all invalidates that will be superflous after this. I'll add that to this PR. 

But yeah I guess we can try it for a while anyway and we'll have to change it up later if it does not work for us.

PR #797, opened 2024-02-09
Bootstrap rust application

junlarsen:
Unsure of how good of an idea this is...

Depends on #795


junlarsen:
Also quite a bit of spaghetti here but it's good enough for an MVP imo

apps/grades-sync/src/main.rs
```diff
@@ -0,0 +1,41 @@
...
+    let subject_repository = SubjectRepositoryImpl::new(&pool);
+    let grade_repository = GradeRepositoryImpl::new(&pool);
+    let job_service = JobServiceImpl::new(
+        &faculty_repository,
+        &department_repository,
+        &subject_repository,
+        &grade_repository,
+    );
+
+    // job_service.perform_faculty_synchronization().await.unwrap();
```
joleeee, line 36: why commented out?

apps/grades-sync/src/grade_repository.rs
```diff
@@ -0,0 +1,270 @@
...
+
+    pub fn to_multiplication_factor(self) -> f32 {
+        match self {
+            SubjectGradingKey::A => 5.0,
+            SubjectGradingKey::B => 4.0,
+            SubjectGradingKey::C => 3.0,
+            SubjectGradingKey::D => 2.0,
+            SubjectGradingKey::E => 1.0,
+            SubjectGradingKey::F => 0.0,
+            x => panic!("invalid grading key: {:?}", x),
```
joleeee, line 51: panicing here is not that great

apps/grades-sync/src/grade_repository.rs
```diff
@@ -0,0 +1,270 @@
...
+        }
```
joleeee, line 52:
```suggestion
    pub fn to_multiplication_factor(self) -> Option<f32> {
        match self {
            SubjectGradingKey::A => Some(5.0),
            SubjectGradingKey::B => Some(4.0),
            SubjectGradingKey::C => Some(3.0),
            SubjectGradingKey::D => Some(2.0),
            SubjectGradingKey::E => Some(1.0),
            SubjectGradingKey::F => Some(0.0),
            _ => None,
        }
```

apps/grades-sync/src/subject_repository.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+impl Subject {
+    pub fn get_next_average(&self, key: SubjectGradingKey, count: i32) -> f32 {
+        if key.is_pass_or_fail_key() {
+            return self.average_grade;
+        }
+
+        let complete_factor = self.average_grade * self.total_students as f32
+            + count as f32 * key.to_multiplication_factor();
```
joleeee, line 29: have to update this also if changing it to be `-> Option<_>`

apps/grades-sync/src/job.rs
```diff
@@ -0,0 +1,203 @@
...
+
+const MAX_TASK_COUNT: usize = 100;
+
+#[async_trait]
+impl<'a> JobService for JobServiceImpl<'a> {
+    async fn perform_faculty_synchronization(&self) -> anyhow::Result<()> {
+        info!("performing faculty synchronization");
+        let departments = get_departments().await?;
+        let department_count = departments.len();
+        let _chunks_count = min(department_count / MAX_TASK_COUNT, 1000);
```
joleeee, line 56: unused variable? if you want to split it into smaller chunks i would just use [std::slice::Chunks](https://doc.rust-lang.org/std/slice/struct.Chunks.html)

apps/grades-sync/src/job.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    async fn synchronize_single_grade(&self, grade: HkdirGrade) -> anyhow::Result<()> {
+        // First we find the matching subject, or else we return early.
+        let subject = match self
+            .subject_repository
+            .find_subject_by_ref_id(grade.subject_code)
+            .await?
+        {
+            Some(subject) => subject,
+            None => return Ok(()),
```
joleeee, line 183: Is it really a success if the grade wasnt found? just wondering

apps/grades-sync/src/job.rs
```diff
@@ -0,0 +1,203 @@
...
+            .await
+            .unwrap();
+        Ok(())
+    }
+
+    async fn synchronize_single_subject(&self, subject: HkdirSubject) -> anyhow::Result<()> {
+        // Forge a slug from the subject name. A lot of the subject code fields from
+        // HKDir have a -1 or another number appended to them, which is not useful
+        // because we're using to seeing 'TDT4120' instead of 'TDT4120-1'.
+        let re = Regex::new("-[A-Za-z0-9]+$").unwrap();
```
joleeee, line 145: this compiles the regex every time this function is called, not sure how much time it takes but could be worth it to move it into a global using `lazy_static` or something

joleeee (changes requested):

joleeee:
So the point of this is to update the grades.no backend db?

apps/grades-sync/src/grade_repository.rs
junlarsen: Yeah, could do. I was also wondering if I just inline it at call-site, since the call site guarantees that it's not G/H

apps/grades-sync/src/job.rs
junlarsen: Yep makes sense

apps/grades-sync/src/job.rs
junlarsen: It means that there's some data inconsistency from the API. Earlier versions have just silent-failed.

I think I'll probably just toss in a warn!() or something

apps/grades-sync/src/grade_repository.rs
junlarsen: Yep. Same case as above. I think I might just inline at call-site

apps/grades-sync/src/main.rs
junlarsen: Forgot to uncomment 💀

junlarsen:
> So the point of this is to update the grades.no backend db?

We are replacing the entire site

apps/grades-sync/src/pg.rs
```diff
@@ -0,0 +1,26 @@
...
+use sqlx::{ConnectOptions, Pool, Postgres};
+use std::time::Duration;
+
+pub type Database = Pool<Postgres>;
+
+pub async fn create_postgres_pool() -> Result<Pool<Postgres>, sqlx::Error> {
+    let database_url =
+        std::env::var("DATABASE_URL").expect("missing DATABASE_URL environment variable");
+    let max_connections = std::env::var("MAX_CONNECTIONS")
+        .unwrap_or_else(|_| "10".to_string())
```
brage-andreas, line 12: Maybe this could be a constant variable instead of a magic number?

jotjern:
Wild pull request, love it

PR #798, opened 2024-02-10
feat/yeet cognito

henrikhorluck:
- Yeet Cognito for Auth0
- Remove auth dispatcher


infra/auth0/inputs.tf
```diff
@@ -0,0 +1,8 @@
+# variable "FEIDE_CLIENT_ID" {
+#   type = string
+# }
+
+# variable "FEIDE_CLIENT_SECRET" {
+#   type      = string
+#   sensitive = true
+# }
```
junlarsen, line 8: These should probably be pulled from Doppler in the future

infra/auth0/main.tf
```diff
@@ -0,0 +1,604 @@
...
+  grant_types                   = ["authorization_code", "refresh_token"]
+  name                          = "Vengeful Vineyard${terraform.workspace == "dev" ? " Dev" : ""}"
+  organization_require_behavior = "no_prompt"
+  is_first_party                = true
+  oidc_conformant               = true
+  
+  jwt_configuration {
+    alg = "RS256"
+  }
+}
```
junlarsen, line 131: We can leave them here for now, but I think that in the future, the other apps themselves should register their own clients

At that point we should probably abstract over with Terraform modules too.

infra/auth0/main.tf
```diff
@@ -0,0 +1,604 @@
...
+
+resource "doppler_secret" "client_ids" {
+  for_each = local.projects
+  project  = each.key
+  config   = terraform.workspace
+  name     = "AUTH0_CLIENT_ID"
+  value    = each.value.client_id
+}
```
junlarsen, line 157: Doing so will avoid requiring having all the configs to run auth0 as well

infra/auth0/ses.tf
```diff
@@ -0,0 +1,44 @@
+
+
+
```
junlarsen, line 3:
```suggestion
```

infra/auth0/ses.tf
```diff
@@ -0,0 +1,44 @@
...
+    resources = [
+      aws_ses_domain_identity.online_ntnu_no.arn
+    ]
+  }
+}
+
+resource "aws_iam_user_policy" "auth0_ses_ro" {
+  user   = aws_iam_user.auth0_ses_emailer.name
+  policy = data.aws_iam_policy_document.auth0_ses_policy.json
+}
```
junlarsen, line 44: newline

infra/auth0/terraform.tf
```diff
@@ -0,0 +1,65 @@
...
+      version = "~> 1.3.0"
+    }
+    auth0 = {
+      source  = "auth0/auth0"
+      version = "~> 1.1.0"
+    }
+  }
+}
+
+variable "DOPPLER_TOKEN_ALL" {
```
junlarsen, line 26: Should stick with terraform namving convention despite it being used as an environment variable
```suggestion
variable "doppler_token_all" {
```

infra/auth0/terraform.tf
```diff
@@ -0,0 +1,65 @@
...
+  }
+}
+
+data "aws_route53_zone" "online" {
+  name = "online.ntnu.no"
+}
+
+data "aws_caller_identity" "current" {}
+
+data "aws_region" "current" {}
```
junlarsen, line 57: Move to data.tf or other file

infra/auth0/text.tf
```diff
@@ -0,0 +1,713 @@
+
+
+resource "auth0_prompt_custom_text" "nb_email_verification" {
```
junlarsen, line 3: In the future I think we'll put this in a json file and for_each this to avoid all the code, but ok for now

packages/db/src/migrations/0023_yeet_cognito.js
```diff
@@ -0,0 +1,19 @@
+import { sql } from "kysely"
+
+/** @param db {import('kysely').Kysely} */
+export async function up(db) {
+  await db.schema
+    .alterTable("ow_user")
+    .dropColumn("cognitoSub")
+    .addColumn("auth0Sub", sql`character varying(50)`, (col) => col.unique())
```
junlarsen, line 8:
```suggestion
    .addColumn("auth0Sub", sql`character varying(50)`, (col) => col.default(sql`null`).unique())
```

packages/db/src/migrations/0023_yeet_cognito.js
```diff
@@ -0,0 +1,19 @@
+import { sql } from "kysely"
```
junlarsen, line 1: Maybe rename this to something more representative than just yeet cognito

junlarsen (changes requested):
Nice, a few comments on the terraform but otherwise looks pretty good

infra/auth0/inputs.tf
henrikhorluck: They are already in doppler for dev

packages/db/src/migrations/0023_yeet_cognito.js
henrikhorluck: I do not care. It represent the action well enough

infra/auth0/terraform.tf
henrikhorluck: Doppler does not support non-uppercasing the env-variable, so I suggest we don't

packages/db/src/migrations/0023_yeet_cognito.js
henrikhorluck: Isn't that already the default? We should add a migration afterwards that add `.notNull()`.

junlarsen: fair enough

infra/auth0/main.tf
```diff
@@ -0,0 +1,613 @@
...
+data "auth0_client" "monoweb_web" {
+  client_id = auth0_client.monoweb_web.client_id
+}
+
+resource "auth0_client" "monoweb_dashboard" {
+  app_type = "regular_web"
+  callbacks = concat(
+    ["https://${terraform.workspace}.dashboard.online.ntnu.no/api/auth/callback/auth0"],
+    {
+      "dev" = ["http://localhost:3000/api/auth/callback/auth0"]
```
henrikskog, line 391:
```suggestion
      "dev" = ["http://localhost:3002/api/auth/callback/auth0"]
```

henrikskog (approved):

PR #806, opened 2024-02-16
Implement landing page in frontend

jotjern:


apps/web/src/components/organisms/EventsPane/MainEventPane.tsx
```diff
@@ -0,0 +1,30 @@
+import { type Event } from "@dotkomonline/types"
+
+function formatDate(date: Date): string {
+  const day = date.getDate()
+  const months = ["jan", "feb", "mar", "apr", "may", "jun", "jul", "aug", "sep", "oct", "nov", "dec"]
+  const month = months[date.getMonth()]
+  const year = date.getFullYear()
+
+  return `${day}. ${month} ${year}`
+}
```
junlarsen, line 10: Would recommend using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat

apps/web/src/components/organisms/EventsPane/index.tsx
```diff
@@ -0,0 +1,3 @@
+import { EventsPane } from "./EventsPane"
+
+export { EventsPane }
```
junlarsen, line 3:
```suggestion
export { EventsPane } from "./EventsPane"
```

packages/core/src/modules/event/event-repository.ts
```diff
@@ -40,6 +41,17 @@ export class EventRepositoryImpl implements EventRepository {
...
+  async getAllOrderedByDate(ascending: boolean, take: number) {
+    const query = this.db
+      .selectFrom("event")
+      .selectAll()
+      .orderBy("start", ascending ? "asc" : "desc")
+      .limit(take)
+
+    const events = await query.execute()
+    return events.map((e) => mapToEvent(e))
+  }
```
junlarsen, line 53: Should take a cursor object here instead and use orderedQuery

packages/gateway-trpc/src/modules/event/event-router.ts
```diff
@@ -47,6 +47,24 @@ export const eventRouter = t.router({
       return event
     }),
 
+  allOrderedByDate: publicProcedure
+    .input(
+      z
+        .object({ take: z.number().default(10), ascending: z.boolean().default(false) })
+        .default({ take: 10, ascending: false })
+    )
```
junlarsen, line 55: Replace with Cursor in repository and use PaginateInputSchema here

junlarsen (commented):
Nice work. I left some code comments, let me know on Slack or something if you want guidance with the cursors

henrikskog:
DOT-314

henrikskog:
https://linear.app/dotkom/issue/DOT-314/create-front-page

PR #807, opened 2024-02-17
Fix navbar hover bug

jotjern:
Fix problem where navbar submenus close before you are able to interact with them when moving the cursor slowly

Uploading Screen Recording 2024-02-17 at 17.31.40.mov…

https://github.com/dotkom/monoweb/assets/16722718/865d940b-2ec8-4a87-8e98-fb13f2c5acfe


henrikskog (approved):
👍👍

junlarsen (approved):

PR #808, opened 2024-02-18
Update README.md

henrikskog:
Make it easier to contribute to monoweb


junlarsen (changes requested):
I think you forgot a commit

junlarsen (approved):

PR #811, opened 2024-02-20
Make core expose contract apis, not direct implementation apis

henrikskog:
Core should only expose the contracts of repositories/services and not the individual implementations. This way we force the contracts to stay updated as well, as previously you could add new methods to a service and get access to it thorugh core without updating interface.

Includes fix for EventCommitteeServiceImpl which was not following the EventCommitteeService interface.

All this PR does is adding the interface types to all repositories/services in core and fixes some issues that arose from this


junlarsen (approved):
Sweet, good catch

PR #817, opened 2024-02-23
Refactor fixture generation

henrikskog:
**Background**
I was creating new fixtures due to some changes that needed to be made to the database schema. I found it very annoying to generate the ulids manually, seeing that the ulids should be absolutely ordered based on insert time. Ended up rewriting such that the ids are generated by the database. 

**Changes**
- Use functions to generate fixtures instead of hard-coded objects so that you can inject FK-IDs from previous fixtures
- some helper functions to make the code a bit neater

### DIFF
--- a/tools/migrator/src/fixture.ts
+++ b/tools/migrator/src/fixture.ts
@@ -6,30 +6,27 @@ import { getCommitteeFixtures } from "./fixtures/committee"
 import { getEventCommitteeFixtures } from "./fixtures/committee-organizer"
 import { getCompanyFixtures } from "./fixtures/company"
 import { getEventFixtures } from "./fixtures/event"
-import { getJobListingFixtures, jobListingLocationLinks, jobListingLocations } from "./fixtures/job-listing"
-import { marks } from "./fixtures/mark"
-import { offlines } from "./fixtures/offline"
-import { personalMarks } from "./fixtures/personal-mark"
-import { products } from "./fixtures/product"
-import { productPaymentProviders } from "./fixtures/product-payment-provider"
-import { users } from "./fixtures/user"
+import {
+  getJobListingFixtures,
+  getJobListingLocationFixtures,
+  getJobListingLocationLinkFixtures,
+} from "./fixtures/job-listing"
+import { getMarkFixtures } from "./fixtures/mark"
+import { getOfflineFixtures } from "./fixtures/offline"
+import { getPersonalMarkFixtures } from "./fixtures/personal-mark"
+import { getProductFixtures } from "./fixtures/product"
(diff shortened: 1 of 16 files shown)


junlarsen:
I'm not sure I understand all the logic here... Are the IDs seeded? We want the ids that fixtures get to be the exact same each time it's ran.

henrikskog:
I did not think that having the id's hard coded is a hard requirement, so I let the database generate them, store the generated ids as the fixtures are being inserted and then use the stored ids when creating fixtures that need the ids for foreign keys.

junlarsen:
> Ah really? Why do we need the ID's to be the same? @junlarsen I guess for testing? I think it is super awkward to maintain the fixtures with hard coded ids.

Re: discussion on Slack, I think it should be fine. With the current setup the fixtures aren't ran in integration tests anyways, so it shouldn't really matter.

tools/migrator/src/fixture.ts
```diff
@@ -1,205 +1,141 @@
...
+  results: { id: string }[]
+) => {
+  results.forEach((result, index) => {
+    if (index < resultIds[key].length) {
+      resultIds[key][index] = result.id
+    }
+  })
+}
+
+const mapId = (results: { id: string }[]) => results.map((res) => res.id)
```
junlarsen, line 37: I suggest we avoid object type literals inline, since they hurt readabilty. I suggest we move it into a type, especially considering it's used other places

```suggestion
type WithIdentifier = { id: string }

const mapId = (results: WithIdentifier[]) => results.map((res) => res.id)
```

tools/migrator/src/fixture.ts
```diff
@@ -1,205 +1,141 @@
...
+  resultIds: InsertedIds,
+  key: T,
+  results: { id: string }[]
+) => {
+  results.forEach((result, index) => {
+    if (index < resultIds[key].length) {
+      resultIds[key][index] = result.id
+    }
+  })
+}
```
junlarsen, line 35: Could you write a small comment on what exactly this does?

junlarsen (changes requested):
Just two small nits, otherwise LGTM

tools/migrator/src/fixture.ts
henrikskog: Ah now I understand what you meant by the logic in your first comment. This is left over code from one of my earlier iterations. Sorry for the confusion!

tools/migrator/src/fixture.ts
henrikskog: What other places were you referring to that the type  should be used?

Btw, I'm using an interface cause our eslint conf perfers interfaces

junlarsen (approved):

PR #820, opened 2024-02-26
Let generic form inputs in form builder interact with global form in useSubmit

henrikskog:
A part of a bigger rewrite of attendance (branch events-POC). However, this change can be evaluated out of the context of that rewrite.

Makes it possible to i.e. clear the form on submit or set errors in the form based on asynchronous operations in the onSubmit. 

Clearing the form is useful in the form for creating new attendance pools as then you can reuse the form and clear the values between each submit. This is backwards compatible with existing usages of the form as it simply adds a new argument to the onSubmit that will be left unused for existing usages of useFormBuilder.

It will be easier to review this PR when the attendance PR is up.


junlarsen (approved):
Makes sense

PR #821, opened 2024-02-26
Replace ESLint & Prettier with Biome

junlarsen:
[biomejs.dev](https://biomejs.dev/) is a modern replacement for both prettier+eslint written in Rust. It's really really fast, like 10x the speed of eslint+prettier and it has a really extensive linting ruleset.

As you can see from the diff, the formatting options are 100% compatible with what we already have.

If we want to, we can also turn on organizeImports which alphabetically sorts the imports. I have not done that yet, since that would dilute the diff with like 100 files changed.

Instead of spending two minutes waiting for lintfix, it now takes 2.5 seconds for all the code

image


henrikhorluck (approved):
:shipit:

henrikhorluck (approved):

brage-andreas (approved):
Good job!

PR #823, opened 2024-02-28
add uri decoder to migration folder path

BillySBarrett:
Migrations won't run when the migrations folder path has any URL encoded characters. This is a problem if the repo is inside a folder with a space character (definitely not me 👀)


henrikskog:
Nice job!

junlarsen (approved):

PR #804, opened 2024-02-14
Implement event attendance MVP

henrikskog:
Lacks functionality for:
- no waitlist domain logic is implemented

New schema for attendance
image
Attendance is completely independant from event. it does not even have a FK to event. this makes it nice to model in the backend.

- Dashboard functionality: 
    - finding users by name and marking as attendance
    - creating new attendance pools (capacity, year restrictions)
    - manage basic stuff for attendasnce like attendance start, end i.e.

- Basic design for event detail page with
    - Functioning attend event button
    - Showing groups the logged in user can attend + all groups

PR #825, opened 2024-03-05
Sync all user data to local db table

henrikskog:
**Changes approach for how we store user data. Takes inspiration from the auth0 documentation**

From [auth0 documentation](https://auth0.com/docs/secure/security-guidance/data-security/user-data-storage#external-database-vs-auth0-data-store):

> For example, you could have a Users table that lists each user authenticated by Auth0. Every time a user logs in, you could search the table for that user. If the user does not exist, you would create a new record. If they do exist, you would update all fields, essentially keeping a local copy of all user data.

**Some more context**
- I deleted the migration that added membership types and just integrated it with the first migration.
- We now sync all user data from auth0 to the local table `ow_user` when a user authenticates (not exactly, it's actually in session callback).

User data from auth0 is synced with the respective row in the ow_user table **if it was last updated more than 24h ago** using the `session` callback. 

This number is somewhat arbitrary, so it should be reconsidered.

The`session`-callback is ran a lot. It runs on refocus (as the `useSession` is probably reran on window refocus). I think it is unneccesary to run the database insert that often.

Other callbacks we could use ([documentation](https://next-auth.js.org/configuration/callbacks#sign-in-callback)):
- `signin`
	- this might be invoked to rarely? not sure
- `redirect`
    - triggered as often as signIn more or less
-  `jwt`
	- From manual testing this is triggered a bunch of times on page load, so I don't think its any better than `session`


packages/auth/src/utils.ts
junlarsen, line 1: Can these go into a service? Auth0SynchronizationService or something

packages/core/src/lib/auth0-repository.ts
```diff
@@ -0,0 +1,54 @@
+import { type OidcUser } from "@dotkomonline/types"
+import { ManagementClient } from "auth0"
+
+export interface Auth0Repository {
+  getBySubject(sub: string): Promise<OidcUser | undefined>
+  getAll(limit: number): Promise<OidcUser[]>
+  searchByFullName(query: string, take: number): Promise<OidcUser[]>
+}
+
+export class Auth0IDPRepositoryImpl implements Auth0Repository {
```
junlarsen, line 10:
```suggestion
export class Auth0RepositoryImpl implements Auth0Repository {
```

packages/core/src/lib/auth0-repository.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    return users.data.map((user) => ({
+      email: user.email,
+      familyName: user.family_name,
+      givenName: user.given_name,
+      subject: user.user_id,
+    }))
+  }
+
+  async getBySubject(IDPuserId: string): Promise<OidcUser | undefined> {
```
junlarsen, line 43:
```suggestion
  async getBySubject(sub: string): Promise<OidcUser | undefined> {
```

junlarsen (changes requested):

packages/auth/src/utils.ts
henrikskog: Yes that makes more sense, good call! Where do you think this service belongs? With auth0-repository in `/core/src/lib`?

tools/migrator/src/fixtures/user.ts
```diff
@@ -5,21 +5,28 @@ export const getUserFixtures: () => Insertable<Database["owUser"]>[] = () => [
...
-    createdAt: new Date("2023-04-30 21:22:17.627253+00"),
-    auth0Sub: "auth0|dddddddd-c376-4e5e-b5fb-4db9bf6cd417",
-    studyYear: 0,
-  },
-  {
-    createdAt: new Date("2023-04-30 21:22:17.627253+00"),
-    auth0Sub: "auth0|7ba93bf4-2e11-40a9-bcd7-9078ee9090bb",
+    createdAt: new Date("2023-04-30 21:22:17.627253+00"), // henrikskog01@gmail.com
+    auth0Sub: "auth0|61ea20bf-6964-4484-b7b8-e81952606973",
+    email: "henrikskog01@gmail.com",
```
henrikhorluck, line 26: IMO these should not be real names/emails, and we should look into how to get a good devX here.

Also these sub's are for dev?

henrikhorluck (changes requested):

packages/auth/src/utils.ts
henrikskog: I put it in `packages/core/src/lib/auth0-synchronization-service.ts` LMK what you think!

packages/auth/src/auth-options.ts
```diff
@@ -43,6 +46,8 @@ export const getAuthOptions = ({
         name: `${profile.given_name} ${profile.family_name}`,
```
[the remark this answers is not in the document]
henrikskog: Nice! I guess some things are left behind from Cognito.

packages/auth/src/utils.ts
```diff
@@ -0,0 +1,48 @@
+import { type ServiceLayer } from "@dotkomonline/core"
+import { User } from "@dotkomonline/types"
+
+type JWTToken = {
```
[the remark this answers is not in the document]
henrikskog: I agree. This is a somewhat lazy solution where I could not access the type in `packages/auth/src/auth-options.ts` due to it being in a module and I was not sure if I could change that or not. I'll take a closer look at this.

packages/db/src/migrations/0001_create_user_and_auth.js
```diff
@@ -5,7 +5,13 @@ import { createTableWithDefaults } from "../utils.js"
...
   const query = sql`CREATE EXTENSION IF NOT EXISTS ulid;`.compile(db)
   await db.executeQuery(query)
-  await createTableWithDefaults("ow_user", { id: true, createdAt: true }, db.schema).execute()
+  await createTableWithDefaults("ow_user", { id: true, createdAt: true, updatedAt: true }, db.schema)
+    .addColumn("email", "varchar(255)", (col) => col.notNull().unique())
+    .addColumn("given_name", "varchar(255)", (col) => col.notNull())
+    .addColumn("family_name", "varchar(255)", (col) => col.notNull())
+    .addColumn("name", "varchar(255)", (col) => col.notNull())
+    .addColumn("study_year", "integer", (col) => col.notNull().defaultTo(-1))
+    .execute()
```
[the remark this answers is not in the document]
henrikskog: Good insights! I would like to discuss this more tomorrow, let's try not to forget

tools/migrator/src/fixtures/user.ts
henrikskog: Yeah this setup is a bit awkward. My sub is for dev at least, I have not checked the others.

packages/core/src/modules/user/user-repository.ts
```diff
@@ -36,6 +36,9 @@ export class UserRepositoryImpl implements UserRepository {
     const user = await this.db
       .updateTable("owUser")
       .set(data)
+      .set({
+        updatedAt: new Date(),
```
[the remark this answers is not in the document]
henrikskog: Good call!

packages/auth/src/utils.ts
junlarsen: Sure, I think that lib folder should probably be deleted at some point but that's ok for now

tools/migrator/src/fixtures/user.ts
henrikskog: I created a new issue for this on Linear. I don't think it's something that should be fixed as a part of this PR.

packages/auth/src/utils.ts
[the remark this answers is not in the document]
henrikskog: I added an explanatory comment. We don't want core to depend on the auth package.

packages/core/src/lib/auth0-synchronization-service.ts
```diff
@@ -0,0 +1,78 @@
...
+export interface Auth0SynchronizationService {
+  /**
+   * Creates a new user in the local database using the provided JWT token.
+   */
+  createNewUser: (token: JWTToken) => Promise<User>
+
+  /**
+   * Synchronizes the provided user with data from Auth0. Only updates user if the user's updatedAt field is more than 1 day ago.
+   */
+  handleSyncUserWithAuth0: (user: User) => Promise<User | undefined>
```
junlarsen, line 26: :nerd_face: :point_up: 
```suggestion
  createUser: (token: JWTToken) => Promise<User>

  /**
   * Synchronizes the provided user with data from Auth0. Only updates user if the user's updatedAt field is more than 1 day ago.
   */
  synchronizeUser: (user: User) => Promise<User | undefined>
```

junlarsen (approved):

packages/core/src/lib/auth0-synchronization-service.ts
```diff
@@ -0,0 +1,74 @@
...
+export interface Auth0SynchronizationService {
+  /**
+   * If no record of the user exists in the local database, save it to the database.
+   */
+  createUser: (token: JWTToken) => Promise<User>
+
+  /**
+   * Synchronize record of user in database with user data from Auth0.
+   */
+  synchronizeUser: (user: User) => Promise<User | undefined>
```
henrikhorluck, line 26:
```suggestion
  synchronizeUser: (user: User) => Promise<User | null>
```

packages/core/src/lib/auth0-synchronization-service.ts
```diff
@@ -0,0 +1,74 @@
...
+        email: idpUser.email,
+        givenName: idpUser.givenName,
+        familyName: idpUser.familyName,
+        name: `${idpUser.givenName} ${idpUser.familyName}`,
+        lastSyncedAt: new Date(),
+      })
+    }
+    return undefined
+  }
+}
```
henrikhorluck, line 74: I would rather think you just perform a sync whenever you get an `id_token` (which is every time a user has to log in I think, and maybe whenever you refresh the access token? I do not know how much nextAuth does though), and that whenever the user wants to change something that we store in auth0, what is stored in auth0 is changed first, then we invalidate the `id_token` and fetch a new one? 

The 1 day is arbitrary and `id_token`s has it's own expiry (which by default is 10 hours). 

This kind of syncing just leads to a one-day gap between changing e.g. givenName and it actually being updated, if we forget to invalidate the id_token?

henrikhorluck (changes requested):

packages/core/src/lib/auth0-repository.ts
```diff
@@ -0,0 +1,21 @@
+import { type OidcUser } from "@dotkomonline/types"
+import { ManagementClient } from "auth0"
+
+export interface Auth0Repository {
+  getBySubject(sub: string): Promise<OidcUser | undefined>
```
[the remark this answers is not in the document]
henrikskog: True, thanks!

packages/auth/src/auth-options.ts
```diff
@@ -52,10 +74,19 @@ export const getAuthOptions = ({
...
-        let user = await core.userService.getUserBySubject(token.sub)
+        const user = await core.userService.getUserBySubject(token.sub)
+
         if (user === undefined) {
-          user = await core.userService.createUser({ auth0Sub: token.sub, studyYear: -1 })
+          // const newUser = await createNewUser(core, token)
+          const newUser = await core.auth0SynchronizationService.createUser(token)
+
+          session.user.id = newUser.id
+          session.sub = token.sub
```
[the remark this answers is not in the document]
henrikskog: What do you think about this issue @junlarsen? In my opinion it is fine to use the sub as an id. Is there any arguments of having a normal ulid id on the table?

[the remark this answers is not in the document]
junlarsen: Either one works, I guess it might be annoying if we ever migrate off Auth0 but I'd say its fine to just use the Auth0 sub here

packages/core/src/lib/auth0-repository.ts
```diff
@@ -0,0 +1,21 @@
...
+  async getBySubject(sub: string): Promise<OidcUser | undefined> {
+    const res = await this.client.users.get({
+      id: sub,
+    })
+    return {
+      email: res.data.email,
+      familyName: res.data.family_name,
+      givenName: res.data.given_name,
+      subject: res.data.user_id,
+    }
```
junlarsen, line 19: Create a zod scema in /packages/types and parse here

packages/core/src/lib/auth0-synchronization-service.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    return this.userService.createUser(userData)
+  }
+
+  // if user.updatedAt is more than 1 day ago, update user
+  async synchronizeUser(user: User) {
+    const oneDay = 1000 * 60 * 60 * 24
+    const oneDayAgo = new Date(Date.now() - oneDay)
+    if (!user.lastSyncedAt || user.lastSyncedAt < oneDayAgo) {
+      console.log("updating user", user.id, user.auth0Sub)
```
junlarsen, line 57: Remove this and replace with a logger from /packages/logger

junlarsen (changes requested):

.env.example
```diff
@@ -8,6 +8,7 @@ WEB_AUTH0_ISSUER=https://dev.id.online.ntnu.no
 GTX_AUTH0_CLIENT_ID=
 GTX_AUTH0_CLIENT_SECRET=
 GTX_AUTH0_ISSUER=https://dev.id.online.ntnu.no
+GTX_AUTH0_DOMAIN=onlineweb.eu.auth0.com
```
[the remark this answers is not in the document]
henrikskog: I can't find this in doppler though?

packages/auth/src/auth-options.ts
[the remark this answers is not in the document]
henrikskog: See https://linear.app/dotkom/issue/DOT-628/refactor-user-to-use-the-id-from-auth0, I think it's better to do this in a seperate PR

packages/core/src/lib/auth0-synchronization-service.ts
henrikskog: > I would rather think you just perform a sync whenever you get an id_token (which is every time a user has to log in I think, and maybe whenever you refresh the access token? I do not know how much nextAuth does though), and that whenever the user wants to change something that we store in auth0, what is stored in auth0 is changed first, then we invalidate the id_token and fetch a new one?

I created this issue for it: https://linear.app/dotkom/issue/DOT-629/find-a-better-solution-for-synchronizing-auth0-user-data-to-ow-user. I'm not quite sure how this is done and I feel blocked by this PR on my #804 PR. In my opinion this is a case of optimization in regards to the amount of database calls (we fetch the user from the database on every `session` callback). I don't see what synchronization problems it leads to. So I would rather just merge this now and find a more optimized solution later.

Thoughts?

> This kind of syncing just leads to a one-day gap between changing e.g. givenName and it actually being updated, if we forget to invalidate the id_token?

When updating givenName we will use the auth0 mgmt api, then sync the user with the database. We would also do it this way if we did not store any user info in the application database. In other words, we will have the same problem with invalidating tokens no matter what right? I don't see how this is relevant to this synchronization approach?

junlarsen: From what I can tell, a potential syncing problem would exist in either case?

I agree that it's probably easiest to merge this now, and worry about moving to synchronizing against id_token in a future PR

packages/core/logger.ts
```diff
@@ -0,0 +1,3 @@
+import { getLogger } from "@dotkomonline/logger"
+
+export const logger = getLogger("core")
```
junlarsen, line 3: This is fine for now, but in the future we want a specific logger in each service where needed, that way it's a lot easier to trace

junlarsen (approved):
- [ ]

junlarsen (approved):
- [ ]

.env.example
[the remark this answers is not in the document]
henrikhorluck: Then I will add it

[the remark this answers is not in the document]
henrikskog: Okay nice

packages/core/src/lib/auth0-repository.ts
```diff
@@ -0,0 +1,36 @@
+import { OidcUser } from "@dotkomonline/types"
+import { ManagementApiError, ManagementClient } from "auth0"
+
+export interface Auth0Repository {
+  getBySubject(sub: string): Promise<OidcUser | null>
+}
+
+export class Auth0RepositoryImpl implements Auth0Repository {
+  constructor(private readonly client: ManagementClient) {}
+  async getBySubject(sub: string) {
```
henrikhorluck, line 10: If we have an access token we can just use `/userinfo`, no need to use the mgmt API.

What is in the `id_token` may have a better schema than the mgmt-api (e.g. `gender` is actually gender instead of being in `user_metadata/gender`, and we _might_ add stuff to the id_token that is not in auth0

packages/db/src/migrations/0001_create_user_and_auth.js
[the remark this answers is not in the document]
henrikhorluck: Changing old migrations is a bad habit to acquire, and if you feel like monoweb is making it hard to properly make migrations we should actually do something about it. 

This change means that everyone that develops on monoweb has to recreate their database-schema and delete all their data which I think is unnecessary, even though it does not cause any harm to any production db this is like throwing a stick into the wheel of other developers

> This way it is clear what the user table includes, rather than having it in many different migration files.

Having fewer migration files is not a solution, people should not be reading migrations to figure out the current db schema

henrikhorluck (changes requested):

packages/core/src/lib/auth0-repository.ts
junlarsen: Probably for fetching attendees on an event?

henrikhorluck: If that user has not logged in, there is no way for the data to go stale?

packages/auth/src/auth-options.ts
```diff
@@ -2,6 +2,24 @@ import { type ServiceLayer } from "@dotkomonline/core"
 import { type DefaultSession, type DefaultUser, type User, type NextAuthOptions } from "next-auth"
 import Auth0Provider from "next-auth/providers/auth0"
 
+interface Auth0TokenClaims {
```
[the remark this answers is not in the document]
henrikskog: Oops! There.

packages/core/src/lib/auth0-synchronization-service.ts
```diff
@@ -0,0 +1,74 @@
+import { User, UserWrite } from "@dotkomonline/types"
+import { UserService } from "../modules/user/user-service"
+import { Auth0Repository } from "./auth0-repository"
+
+// Id token returned from Auth0. We don't want core to depend on next-auth, so we duplicate the type here.
+type JWTToken = {
```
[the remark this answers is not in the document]
henrikskog: Mixed up my branches while revieweing, made the change on the other PR im working on. There it's fixed. Sorry about that.

packages/db/src/migrations/0001_create_user_and_auth.js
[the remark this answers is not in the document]
henrikskog: I researched this more now and I realize that this change will indeed be very annoying for other people that are using their own database. 

Am I right in this thinking:

In this PR I am adding fields to the `ow_user` with the `not null` restriction. E.g. `name`.

The proper way of doing this would be to:
1. create new migration
2. create new fields WITHOUT not null restriction
3. modify all rows with values for the new fields that are not null
4. modify fields to have not null restriction

I have not been working with migrations before so for me this seems like a tideous process. However, maybe it's not, I don't know.

My workflow when creating these new migrations have been to just delete the database and create it before running migrations. But I understand this is not something people have set up.

packages/core/src/lib/auth0-repository.ts
henrikskog: What about this scenario:
1. User goes to their profile on our website and change their email
2. monoweb uses management api to update email in auth0

Now it is important that synchronizations after this actually use the new user state and not some stale data from id tokens. 

For the id tokens to contain the new user data, the user has to authorize again. Hitting the /userinfo endpoint will return an id token with stale data (old email before patch). Read [this](https://community.auth0.com/t/auth0-returns-a-stale-userinfo/13030/4)

If we just use the management api we can make sure we never use stale data from auth0 when synchronizing.

Thoughts?

henrikhorluck: Ah, that is kinda unexpected but also understandable. Does the id_token update if you refresh the access_token?

henrikskog: I don't know. I suggest we use the management api for now.

packages/db/src/migrations/0001_create_user_and_auth.js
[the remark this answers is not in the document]
henrikhorluck: This is a pain to deal with if anything goes wrong, IMO devs _should just need to apply a migration_. It is valuable to have an examples of proper migration for later devs, and you benefit from learning how to do this properly. This kind of migration will be needed later as well.

I leave it to your discretion to find a suitable solution

packages/db/src/db.generated.d.ts
```diff
@@ -192,8 +192,12 @@ export interface Offline {
 export interface OwUser {
   auth0Sub: string | null
   createdAt: Generated<Timestamp>
+  email: string
   id: Generated<string>
+  lastSyncedAt: Timestamp | null
```
henrikhorluck, line 197:
```suggestion
  lastSyncedAt: Timestamp
```
This should just be default now?

packages/db/src/migrations/0024_user.js
```diff
@@ -0,0 +1,43 @@
...
+    .addColumn("email", "varchar(255)")
+    .addColumn("name", "varchar(255)")
+    .addColumn("updated_at", "timestamptz", (col) => col.defaultTo(sql`now()`).notNull())
+    .addColumn("last_synced_at", "timestamp with time zone", (col) => col.defaultTo(null))
+    .execute()
+
+  // populate data for new columns
+  const users = await db.selectFrom("ow_user").selectAll().execute()
+  for (const user of users) {
+    const email = `${user.id}@gmail.com`
```
henrikhorluck, line 16:
```suggestion
    const email = `${user.id}@invalid.local`
```

henrikhorluck (approved):

packages/db/src/db.generated.d.ts
henrikskog: Yea agree, good suggestion.

packages/db/src/migrations/0024_user.js
henrikskog: smart!

PR #826, opened 2024-03-07
Feature/dot 354 design and implement offline component

hanlun0804:
Add an offline component for monoweb's homepage. 

Closes: #...

## Checklist

- [ ] I have written tests

## Changelog

- Implementet component from scratch
- Use position on the maps to make it scroll smoothly

## How to test

Place component a convenient place in monoweb and use an array of links to images as parameters. Pass this as both "images" and "pdfs". Component will use an array of links to the frontpage as images and an array of links to the pdfs in the future, when these are possible to access.


henrikhorluck (commented):
I think that as a reviewer I should not have to write any code, could you please either add the component somewhere it is used?

For example: this could be moved to the ui-package so that it is shown in ladle (localhost:61000), and it should ideally have some data-fixtures available to check it working

@junlarsen probably has better control of where to place this

PR #829, opened 2024-03-09
Improve performance in integration tests

henrikskog:
Set up _one_ container at the start of running tests. Instead of starting new containers for each test, do this:

```
beforeEach(async () => {
  const env = createEnvironment()
  await runMigrations(env)
})

afterEach(async () => {
  const env = createEnvironment()
  await resetTestDatabase(container, env)
})
```

**Note**:
I added the `withReuse()` option to the `PostgreSqlContainer`. I __think__ this should run fine in ci but I'm not 100% sure.

### Benchmarks

**Before:**
```
 ✓ src/modules/user/__test__/user-service.e2e-spec.ts (4) 35020ms
   ✓ users (4) 35020ms
     ✓ can create new users 9889ms
     ✓ will not allow two users the same subject 8424ms
     ✓ will find users by their user id 8364ms
     ✓ can update users given their id 8342ms
```

**After:**
Without reusing container
```
✓ src/modules/user/__test__/user-service.e2e-spec.ts (4) 13469ms
   ✓ users (4) 4478ms
     ✓ can create new users 1239ms
     ✓ will not allow two users the same subject 1089ms
     ✓ will find users by their user id 1084ms
     ✓ can update users given their id 1064ms
```

    
With reusing container
```
 ✓ src/modules/user/__test__/user-service.e2e-spec.ts (4) 4924ms
   ✓ users (4) 4098ms
     ✓ can create new users 838ms
     ✓ will not allow two users the same subject 1098ms
     ✓ will find users by their user id 1105ms
     ✓ can update users given their id 1056ms
```

### DIFF
--- a/packages/core/vitest-integration.setup.ts
+++ b/packages/core/vitest-integration.setup.ts
@@ -37,28 +37,27 @@ async function runMigrations(env: Environment) {
 
 async function resetTestDatabase(container: StartedPostgreSqlContainer, env: Environment) {
   try {
+    // Create client for the default database and use it to drop and recreate the test database
+    const dbString = container.getConnectionUri().replace("/main", "/postgres")
+    const kysely = createKysely({
+      ...env,
+      DATABASE_URL: dbString,
+    })
 
-  // Create client for the default database and use it to drop and recreate the test database
-  const dbString = container.getConnectionUri().replace("/main", "/postgres")
-  const kysely = createKysely({
-    ...env,
-    DATABASE_URL: dbString,
-  })
+    const deleteQuery = sql`drop database if exists ${sql.ref(testDBConfig.database)}`.compile(kysely)
+    const createQuery = sql`create database ${sql.ref(testDBConfig.database)}`.compile(kysely)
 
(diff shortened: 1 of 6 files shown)


packages/core/vitest-integration.setup.ts
```diff
@@ -1,28 +1,67 @@
...
+
+  containers.push(container)
   process.env.DATABASE_URL = container.getConnectionUri()
-  const env = createEnvironment()
-  const kysely = createKysely(env)
+}, 30000)
+
+beforeEach(async () => {
+  env = createEnvironment()
+  kysely = createKysely(env)
```
henrikhorluck, line 35: Each test should probably be run within a transaction for isolation?

This is out-of-the-box in [django](https://docs.djangoproject.com/en/5.0/topics/testing/tools/#transactiontestcase)

henrikskog: I'm not sure about this. Could you explain some more on what you mean by isolation?

With this setup the database is deleted and recreated between each test.

junlarsen (commented):
Is there a reason for using separate databses vs creating+dropping the public schema?

junlarsen (approved):

PR #831, opened 2024-03-09
Clean up and reduce boilerplate in test setup

junlarsen:
**Note: this is a PR against the branch of #829 **

Removes the need to elide dbName into variable, making it less error prone. Also does sharing of Kysely instances

### DIFF
--- a/packages/core/vitest-integration.setup.ts
+++ b/packages/core/vitest-integration.setup.ts
@@ -17,7 +17,7 @@ async function runMigrations(env: Environment, dbName: string) {
 function createKyselyForDatabase(env: Environment, dbName: string): Kysely<Database> {
   return createKysely({
     ...env,
-    DATABASE_URL: `postres://local:local@localhost:${container.getFirstMappedPort()}/${dbName}`,
+    DATABASE_URL: `postres://local:local@${container.getHost()}:${container.getFirstMappedPort()}/${dbName}`,
   })
 }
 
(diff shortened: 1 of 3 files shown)


packages/core/vitest-integration.setup.ts
```diff
@@ -1,81 +1,68 @@
...
-  }@localhost:${container.getFirstMappedPort()}/${dbName}`
-}
-
-export function getTestDb(env: Environment, dbName: string): Kysely<Database> {
-  const url = buildDbUrl(dbName)
+function createKyselyForDatabase(env: Environment, dbName: string): Kysely<Database> {
   return createKysely({
     ...env,
-    DATABASE_URL: url,
+    DATABASE_URL: `postres://local:local@localhost:${container.getFirstMappedPort()}/${dbName}`,
```
henrikskog, line 20:
```suggestion
    DATABASE_URL: `postres://local:local@${container.getHost()}:${container.getFirstMappedPort()}/${dbName}`,
```

https://java.testcontainers.org/features/networking/
> Getting the container host
> When running with a local Docker daemon, exposed ports will usually be reachable on localhost. However, in some CI environments they may instead be reachable on a different host.

> As such, Testcontainers provides a convenience method to obtain an address on which the container should be reachable from the host machine.

henrikskog (approved):
Really nice cleanup!

PR #832, opened 2024-03-10
wipe database as first step when running fixtures script

henrikskog:
When running `migrate-with-fixtures` you want to run this on a clean slate database, or else you will get conflicts with existing data. This adds an option to the migrator to wipe the database as the first thing that happens, and changes the `migrate-with-fixtures` script to use it.


tools/migrator/src/index.ts
```diff
@@ -74,8 +81,8 @@ program
     }
 
     if (res.error) {
-      logger.warn("Error while running migrations:")
-      logger.warn(JSON.stringify(res.error))
+      logger.warn("Error while running migrations")
+      console.dir(res.error, { depth: null }) // prints stack trace
```
henrikskog, line 85: `logger.warn(JSON.stringify(res.error))` does not print stack trace and is super annoying when getting errors

henrikhorluck (changes requested):
I would *not* expect a migration command to just casually drop my database, and in particular not without a warning. Fixtures is just adding new data

henrikskog:
Yeah I see your point. I guess I am a bit biased in how I have only used this command in combination with deleting my database, but I strongly agree with how deleting the database should never come as a surprise when running the script. What do you think about it now?

junlarsen (commented):
Can't we just do `onConflict((eb) => eb.doNotihng()))` or whatever it is?

junlarsen:
> Can't we just do `onConflict((eb) => eb.doNotihng()))` or whatever it is?

Alternatively, update the conflicting fields to take the new value?

henrikskog:
Aaaa that's what that was for! Yeah we could do that. I removed all of that in https://github.com/dotkom/monoweb/pull/817 oops 😅 

However, as this just adds a new script I don't think there is any harm in merging this?

henrikhorluck (changes requested):
> When running migrate-with-fixtures you want to run this on a clean slate database, or else you will get conflicts with existing data.

No, you don't need to have a clean slate? Isn't that only a thing you would want because you are editing old migrations? The migrations should be able to run regardless of what data you currently have, assuming they were properly written?

junlarsen:
> Aaaa that's what that was for! Yeah we could do that. I removed all of that in #817 oops 😅
> 
> However, as this just adds a new script I don't think there is any harm in merging this?

I think I agree with @henrikhorluck I don't think running any fixtures should ever drop the entire database, even if it's just adding a new command

tools/migrator/src/fixture.ts
```diff
@@ -33,83 +33,95 @@ export const runFixtures = async () => {
 
   insertedIds.owUser = await db //
     .insertInto("owUser")
+    .onConflict((conflict) => conflict.doNothing())
```
junlarsen, line 36: very very nitpicky, but can we name this eb to keep consistent with kysely?

junlarsen:
I'm happy to merge in the onConflict changes though!

henrikskog:
Fair enough guys :( The problem is really actually solved by just adding the 
```
    .onConflict((conflict) => conflict.doNothing())
```

I'll create a new PR and close this

tools/migrator/src/fixture.ts
henrikhorluck: Why would your fixture loading create conflicts? If the fixtures are not up to date that should be a part of our test-suite?

PR #835, opened 2024-03-11
Make tRPC context authenticate using Auth0 JWT

junlarsen:
Since we are working on migrating the tRPC server out of the web Next.js app, we want to move away from using Next-Auth sessions to authenticate the currently logged in user.

This patch refactors both Next.js apps to work following the backend-for-frontend paradigm with a proxy endpoint which attaches a Bearer token with the currently signed in user's access token from Auth0.

If the access token is expired, the code will attempt to get a new token using the refresh token. If that also fails, it returns 401


junlarsen:
> Great work!
> 
> I'm a bit confused about the backend-for-frontend part. Note that I have only read [this](https://philcalcado.com/2015/09/18/the_back_end_for_front_end_pattern_bff.html) article on the pattern, so I am by no means an expert here.
> 
> My understanding of the pattern after reading the article is that you build a seperate api for each application (web and dashboard) to allow for creating application spesific routes, without cluttering the common API. In this PR, the applications still use the same API right? They would stille use the same appRouter?
> 
> Or is the plan to in the future split appRouter into application spesific routers?
> 
> Is this something you are thinking of employing in the future?

Typically you would build a backend for each separate application, think web/ios/android, but since we have two web apps, and tRPC is our internal API we will only have one backend.

You can imagine the API Gateway we are planning for Appkom to also act as a BFF (just for all external apps)

There will only ever be one tRPC AppRouter

apps/dashboard/src/middleware.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const middleware = createMiddleware({
+  issuer: env.GTX_AUTH0_ISSUER,
+  clientId: env.GTX_AUTH0_CLIENT_ID,
+  clientSecret: env.GTX_AUTH0_CLIENT_SECRET,
+  gatewayUrl: env.NEXT_PUBLIC_NODE_ENV === "production" ? "https://web.online.ntnu.no" : "http://localhost:3000",
+})
+
+export const config = {
+  matcher: ["/api/proxy/trpc/:path*"],
```
henrikhorluck, line 12: Why is the route itself called proxy? This leaks implementation details that the client does not care about

packages/auth/src/auth-options.ts
```diff
@@ -37,6 +38,13 @@ declare module "next-auth" {
   }
 }
 
+declare module "next-auth/jwt" {
+  interface JWT extends Record<string, unknown>, DefaultJWT {
```
henrikhorluck, line 42:
```suggestion
  interface JWT extends Record<string, unknown>, DefaultJWT {
```
This is the nextAuth session-JWT? _Not_ one from auth0?

packages/jwt-crypto/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+    const issuerWithoutTrailingSlash = issuer.replace(/\/$/, "")
+    this.jwksUrl = new URL(`${issuerWithoutTrailingSlash}/.well-known/jwks.json`)
+  }
+
+  /**
+   * Verify a JWT token using the specified issuer's JWKS
+   */
+  public async verify(token: string) {
+    this.jwks ??= createRemoteJWKSet(this.jwksUrl)
+    return jwtVerify(token, this.jwks, {
```
henrikhorluck, line 18: https://auth0.com/docs/secure/tokens/access-tokens/validate-access-tokens

This is missing audience validation, it should contain `https://online.ntnu.no` (but we currently only have one api/default audience)
We should have scope-validation but that is a TODO

packages/jwt-nextjs/src/index.ts
```diff
@@ -0,0 +1,91 @@
...
+          clientId,
+          clientSecret,
+        })
+        req.headers.set("Authorization", `Bearer ${accessToken}`)
+        return NextResponse.rewrite(endpoint, {
+          request: {
+            headers: req.headers,
+          },
+        })
+      }
```
henrikhorluck, line 83: https://auth0.com/docs/secure/tokens/refresh-tokens/refresh-token-rotation 

I think this applies even if Monoweb is not conf'd as a SPA, as it is an example in the [OAuth2 RFC Refresh Tokens Security Considerations](https://datatracker.ietf.org/doc/html/rfc6749#section-10.4) 

By requesting a new access token you also get a new refresh-token, which invalidates the refresh-token you used. This appears to not store the new request-token in the nextAuth session, so this does not work as you want, and users will get logged out in the case the code runs since the nextAuth-session will have a outdated access-token and an invalid refresh-token.

I do not think we should be dealing with refresh tokens here, does nextAuth not already handle this?

henrikhorluck (changes requested):

apps/dashboard/src/middleware.ts
junlarsen: ah right now it's just called that because otherwise it would clash in the web app, I can change that

packages/auth/src/auth-options.ts
junlarsen: Yes that's correct

packages/jwt-nextjs/src/index.ts
junlarsen: Yeah you are probably right that this isn't the place. I will do what https://next-auth.js.org/v3/tutorials/refresh-token-rotation explains

PR #836, opened 2024-03-11
Document which applications run where

junlarsen:
Just so it's clear and easy to see :)


henrikskog (approved):
Nice!

junlarsen:
Also fixes https://linear.app/dotkom/issue/DOT-464/inform-developer-about-problems-with-occupied-ports

PR #837, opened 2024-03-11
Remove sanity application

junlarsen:
This thing has not been touched in 10+ months safe to say nobody uses it. Removing the codebase (at least for now, maybe we decide to use Sanity and bring it back?) removes 265 dependencies, so I think it's probably worth to delete for now

image

Also see https://linear.app/dotkom/issue/DOT-648/explore-image-hosting-solutions


henrikskog (approved):
Pragmatic solution, nice!

PR #838, opened 2024-03-12
Refactor all external sdks to be injected through dependency injection

junlarsen:
Previously, the S3 repository and Stripe code both created the SDKs on their own. These should instead be provided through dependency injection.

Also removes the /lib favor in core of one named external.


henrikskog (approved):
Nice!

PR #840, opened 2024-03-12
Document how we've used domain driven design

junlarsen:
Quite a bit of yapping but it should give beginners an idea of why we do it, how it's useful, and how we implement it


docs/domain-driven-design.md
```diff
@@ -0,0 +1,118 @@
+# Domain Driven Design
+
+The codebase and project attempts to follow the principles of Domain Driven Design (DDD) as much as possible, and
+wherever it makes sense. DDD is a software development approach that focuses on the domain, the problem that the
+software is trying to solve, and the business logic that drives the problem.
+
+It is a way of thinking and a set of priorities, aimed at accelerating software projects that have to deal with
+complex domains that are constantly changing.
+
+- [Principles of DDD]
```
hanlun0804, line 10: Should these be links to the respective headers?

hanlun0804:
I think it was a nice introduction, some yapping is nice when you go from not knowing what it is :)

docs/domain-driven-design.md
hanlun0804: In that case:
- [Principles of DDD](#principles-of-ddd)
- [Hexagonal Architecture](#hexagonal-architecture)
- [Implemented in code](#implemented-in-code)

henrikskog:
Looking good! I would like a lot more examples, but this is something we can iterate on going forward

henrikskog:
It would be nice to include some examples of the patterns we follow in repositories and services. For example, when a `delete` method in a repository returns `null`, it means that there was no such resource to delete. This sort of stuff is important to document I think

junlarsen:
I agree we should add stuff about "return null" and other paradigms we've used, but I think they are probably better fit in another document

docs/domain-driven-design.md
junlarsen: Yep my bad

PR #842, opened 2024-03-13
Add isActive functionality to interest groups in monoweb admin

hanlun0804:
Closes: #...

## Checklist

- [ ] I have written tests

## Changelog

- edit-card.tsx: Add checkbox to set as active/not active
- 0024_create_interest_group.js: Add column for isActive
- packages/types/src/interest-group.ts: Add isActive to schema
- tools/migrator/src/fixtures/interest-group.ts: Add fixture for testing

## How to test

A consise description of how to test the newly implemented functionality.


apps/dashboard/src/app/(dashboard)/interest-group/[id]/edit-card.tsx
```diff
@@ -33,11 +33,23 @@ export const InterestGroupEditCard: FC = () => {
...
+        <Checkbox label="Aktiv?" onChange={ () =>
+            edit.mutate({
+              id: interestGroup.id,
+              values: {
+                name: interestGroup.name,
+                description: interestGroup.description,
+                link: interestGroup.link,
+                isActive: !interestGroup.isActive,
+              }
+            })} checked={interestGroup.isActive} />
```
Comment (BillySBarrett) [inline apps/dashboard/src/app/(dashboard)/interest-group/[id]/edit-card.tsx:51]: you should use spread elements to avoid having to define all the mutate values

`                    ...interestGroup,
                    values: {
                        ...interestGroup,
                        isActive: !interestGroup.isActive,
                    },
`

BillySBarrett (changes requested):
Good job, just a couple of suggestions! 👍

apps/dashboard/src/app/(dashboard)/interest-group/[id]/edit-card.tsx
Comment (hanlun0804) [reply apps/dashboard/src/app/(dashboard)/interest-group/[id]/edit-card.tsx]: Didn't know that was an option, thanks :)

apps/dashboard/src/app/(dashboard)/interest-group/[id]/edit-card.tsx
```diff
@@ -33,11 +33,22 @@ export const InterestGroupEditCard: FC = () => {
...
+        </Button>
+        <Checkbox
+          label="Aktiv?"
+          onChange={() =>
+            edit.mutate({
+              ...interestGroup, values: { ...interestGroup, isActive: !interestGroup.isActive, },
+            })
+          }
+          checked={interestGroup.isActive}
+        />
```
Comment (BillySBarrett) [inline apps/dashboard/src/app/(dashboard)/interest-group/[id]/edit-card.tsx:50]: I am now a little unsure if this is needed. It may be more practical to add  a useCheckboxInput in the useInterestGroupWriteForm instead such that the update button activates the edit mutation

BillySBarrett (changes requested):

pnpm-lock.yaml
henrikhorluck, line 1: This PR appears to be something that should not touch pnpm-lock at all?

packages/db/src/migrations/0024_create_interest_group.js
henrikhorluck, line 1: This should add a new migration not modify an existing one. Also this should probably just have a default of true?

See https://github.com/dotkom/monoweb/pull/825#discussion_r1513549976 for an earlier discussion on this

apps/dashboard/src/app/(dashboard)/interest-group/[id]/edit-card.tsx
```diff
@@ -33,11 +33,22 @@ export const InterestGroupEditCard: FC = () => {
...
   return (
-    <div>
+    <Stack gap="xs">
       <FormComponent />
-      <Button variant="outline" color="red" style={{ marginTop: "1rem" }} onClick={open}>
-        <Icon icon="tabler:trash" />
-      </Button>
-    </div>
+      <div style={{ display: "flex", flexWrap: "wrap", alignItems: "center" }}>
+        <Button variant="outline" color="red" style={{ width: "55px", marginRight: "24px" }} onClick={open}>
```
Comment (henrikhorluck) [inline apps/dashboard/src/app/(dashboard)/interest-group/[id]/edit-card.tsx:39]: This is unrelated but why are we using inline styling in the dashboard but tailwind elsewhere??

henrikhorluck (changes requested):
Requesting changes since this should add a new migration, otherwise LGTM

pnpm-lock.yaml
hanlun0804: Agreed, never worked with this stuff before, so I don't really know what happened. Will ask about it in the next meeting.

apps/dashboard/src/app/(dashboard)/interest-group/[id]/edit-card.tsx
Comment (henrikskog) [reply apps/dashboard/src/app/(dashboard)/interest-group/[id]/edit-card.tsx:39]: Mantine components have props that work very similarly to tailwind, see https://mantine.dev/styles/style-props/. You should use these instead of the inline styling 👍

apps/dashboard/src/app/(dashboard)/interest-group/[id]/edit-card.tsx
Comment (Terbau) [reply apps/dashboard/src/app/(dashboard)/interest-group/[id]/edit-card.tsx]: Fixed

apps/dashboard/src/app/(dashboard)/interest-group/[id]/edit-card.tsx
Comment (Terbau) [reply apps/dashboard/src/app/(dashboard)/interest-group/[id]/edit-card.tsx]: Fixed

packages/db/src/migrations/0024_create_interest_group.js
Terbau: Fixed

PR #844, opened 2024-03-17
Create central registry for supported error detail types

henrikskog:
**NOTE**: This is a PR against the `feat/proper-error-handling` branch, used in [this pr](https://github.com/dotkom/monoweb/pull/839)

Creates a registry for the problem details types the application can return.

Uses `detail` member for spesific error details instead of `title` member.


packages/core/src/utils/http-response-constants.ts
henrikskog, line 1: I considered using https://www.npmjs.com/package/http-status-codes, but I want the links to the RFC for using for the `type`. I think this is clean enough, and this stuff is of course very stable so it is not something we will have to maintain afaik.

Do you have suggestions for the name of the file and the location which it is placed atm?

packages/core/src/error.ts
```diff
@@ -13,14 +15,45 @@ export class ApplicationError extends Error {
...
   ) {
-    super(title)
+    super(detail || title)
   }
 }
 
 export class IllegalStateError extends ApplicationError {
   constructor(description: string) {
-    super("/problem/illegal-state", 500, `Illegal state reached: ${description}`)
+    const { type, code, title } = serverErrorResponse.InternalServerError
```
junlarsen, line 27: I would much rather overload the constructor of ApplicationError than have to destructure every single time

packages/core/src/error.ts
```diff
@@ -13,14 +15,45 @@ export class ApplicationError extends Error {
   constructor(
     public readonly type: string,
     public readonly status: number,
-    public readonly title: string
+    public readonly title: string,
+    public readonly detail?: string
   ) {
-    super(title)
+    super(detail || title)
```
junlarsen, line 21: Use nullish coalesce
```suggestion
    super(detail ?? title)
```

packages/core/src/modules/event/attendance-service.ts
```diff
@@ -1,6 +1,6 @@
 import { type AttendanceId, type Attendee, type EventId, type UserId } from "@dotkomonline/types"
 import { type AttendanceRepository } from "./attendance-repository"
-import { AttendeeNotFoundError } from "./attendee-error"
+import { EventNotFoundError } from "./event-error"
```
junlarsen, line 3: Can you explain why EventNotFoundError is used here? The idea is to encode the specific cases. Any missing entity in the Event domain should not throw EventNotFoundError, but the specific type that is missing

packages/core/src/utils/http-response-constants.ts
```diff
@@ -0,0 +1,227 @@
...
+    title: "Temporary Redirect",
+    type: `${base}-307-temporary-redirect`,
+  },
+  PermanentRedirect: {
+    code: 308,
+    title: "Permanent Redirect",
+    type: `${base}-308-permanent-redirect`,
+  },
+}
+export const clientErrorResponse = {
```
junlarsen, line 58: SCREAMING_SNAKE_CASE for global constants

packages/core/src/error.ts
```diff
@@ -13,14 +15,45 @@ export class ApplicationError extends Error {
...
 
 export class IllegalStateError extends ApplicationError {
   constructor(description: string) {
-    super("/problem/illegal-state", 500, `Illegal state reached: ${description}`)
+    const { type, code, title } = serverErrorResponse.InternalServerError
+    super(type, code, title, `Illegal state reached: ${description}`)
+  }
+}
+
+export class InternalServerError extends ApplicationError {
```
junlarsen, line 32: I don't think I agree with these types in the middle. I think each error should probably just pass the ServerErrorRespose.X things manually in the constructor

For example, if ApplicationError takes the entire object

```ts
class EventNotFoundError extends ApplicationError {
  constructor(id: string) {
    super(ServerErrorResponse.InternalServerError, `Event with ID:${id} not found`)
  }
}

**Review by junlarsen — CHANGES_REQUESTED** [review]
I'm not sure I fully agree with the changes here

I think we can give a lot better human readable messages than "Unprocessable Entity" and so on.

File: packages/core/src/error.ts (hunk 5 above)
Comment (junlarsen) [reply packages/core/src/error.ts:32]: This of course requires changing the constructor of ApplicationError, but I think in the above suggestion, the code is very readable and understandable

File: packages/core/src/error.ts (hunk 1 above)
Comment (henrikskog) [reply packages/core/src/error.ts:27]: I had not thought about that, I think I like that better too. I'll try it out.

File: packages/core/src/error.ts (hunk 2 above)
Comment (henrikskog) [reply packages/core/src/error.ts:21]: Agreed 👍

**Comment by henrikskog** [comment]
> I'm not sure I fully agree with the changes here
> 
> I think we can give a lot better human readable messages than "Unprocessable Entity" and so on.

With these changes the more spesific messages are given in `detail`, however each error is wrapped in the adequate generic http error, e.g. "Unrpocessable Entity" for POST requests that do not match validation checks.


I made quite a few changes after your last review. Interested to hear what you think.

File: packages/core/src/error.ts [hunk 6]
```diff
@@ -1,26 +1,53 @@
...
+/**
+ * Represents a problem details type.
+ *
+ * This implementation does not yet support:
+ * - instance member
+ * - extension members
+ *
+ * @see https://datatracker.ietf.org/doc/html/rfc9457#name-the-problem-details-json-ob
+ */
+export interface ProblemDetailsType {
```
henrikskog, line 12: I am not sure if this belongs here or in `packages/core/src/problem-details-registry.ts`

junlarsen:
```suggestion
export type ProblemDetail = Pick<ApplicatioNError, 'type' | 'title' | 'status'>
```

If something is called `WhateverType` then the Type part can probably be removed.

packages/core/src/problem-details-registry.ts
junlarsen, line 1: Maybe just call this file `http-status.ts`? I don't quite like the "Problem Details" name

junlarsen (changes requested):
Two small nits, otherwise LGTM

packages/core/src/problem-details-registry.ts
henrikskog: I get where youre coming from, but I think `http-status` might be a bit misleading? This is not just a list of a 1-1 mapping from http status codes to titles. We will have multiple Problem Details types for many status codes.

"Problem details" is the terminology used in rfc9457, I would think it is smart to use well established terms?

What about `http-problem-details.ts`, skipping the registry thing which may be a bit weird

Wh

junlarsen: Mind calling it http-problem-detail in singular? Otherwise lgtm

henrikskog: Sure, but then I'll go over other code as well and change as I have been trying to be consistent in using it in plural.

henrikskog:
I mean there are many problem details in the file, so I think `packages/core/src/http-problem-details.ts` is good.

PR #846, opened 2024-03-17
Fix auth0 synchronization not working

henrikskog:
Fixes a really ugly bug caused by bad code. I typed the domain-map-function argument as `unknown` for the auth0 repository user fetch function, and passed in the raw response from the auth0 sdk to map it to our domain user type. However, the sdk returns the data on the `.data` of the response. This is not catched by typescript, as the parameter is typed `unknown`. Note to self that `unknown` is bad

This made the call to Auth0Repository in the synchronization call fail, which means you could not sign up as you would never be able to sync down the user data to the user table, which is the one the rest of the application uses.

I also add some comments here


packages/auth/src/auth-options.ts
```diff
@@ -84,7 +84,7 @@ export const getAuthOptions = ({
           return session
         }
 
-        await core.auth0SynchronizationService.synchronizeUser(user)
+        await core.auth0SynchronizationService.handleUserSync(user)
```
henrikskog, line 87: I think this makes more sense, as it most often does not do any synchronization?

You can debate it though, because from the perspective of the contract, the method just makes sure that the user is synchronized between auth0 and the database.

packages/core/src/modules/external/auth0-repository.ts
```diff
@@ -15,7 +20,11 @@ export class Auth0RepositoryImpl implements Auth0Repository {
         id: sub,
       })
 
-      return mapToOidcUser(user)
+      return mapToOidcUser({
```
henrikskog, line 23: this is the bug that this PR fixes, which is the main point of the PR

packages/core/src/modules/external/auth0-repository.ts
```diff
@@ -15,7 +20,11 @@ export class Auth0RepositoryImpl implements Auth0Repository {
         id: sub,
       })
 
-      return mapToOidcUser(user)
```
henrikskog, line 18: this is the bug that this PR fixes, which is the main point of the PR

henrikhorluck (changes requested):
All bug fixes should have tests. Might make sense to let some tests go against our dev-tenant? Either that or store example responses and mock it

henrikskog:
Interesting take. When changing the type from unknown the previous code won't compile though.

junlarsen (approved):
Normally typing as unknown is fine as tests are supposed to deal with any exceptions, but I suppose this is fine

I don't really see the reason behind renaming, but it's ok

junlarsen:
Agree with @henrikhorluck that tests should be written

henrikskog:
I have now written a e2e test for the sync service. I have had quite some trouble mocking the auth0 sdk.

I attempted to mock just the ManagementClient from the auth0 module using `vi.mock` like the first example in [the docs](https://vitest.dev/api/vi.html#vi-mock), but I ran into some weird issues. I could verify that client was being mocked, however inside the `createServiceLayer` call, the mocked `users.get` on the mock only returned undefined, even though the mock was being applied. Super weird and super annoying. 

I think it may have been caused by [this](https://vitest.dev/guide/mocking.html#mocking-pitfalls) but I'm not sure.

 I ended up using https://github.com/eratio08/vitest-mock-extended after reading [this issue](https://github.com/vitest-dev/vitest/issues/351) which was asking for just what I wanted.

Ideally this would work without `vitest-mock-extended`, and this is probably something that should be taken a look at.  If one of you wants to take a look at it I would greatly appreciate it.

Note that I also cleaned up the `handleUserSync` function. Some of the sync logic was being applied in `packages/auth/src/auth-options.ts`, so I moved everything to the service.

henrikskog:
I'm going to merge this as it is fixing an important bug. If you find anything you want me to change, LMK and I'll open a new PR for it @junlarsen @henrikhorluck

PR #848, opened 2024-03-19
feature/dot 685 move extras from event to attendance

henrikskog:
**This is finished but it should not be merged before #804 is merged with main**

Move extras from event to attendance.

Also makes it impossible to change extra configuration after registration start. This is because we are not handling it. If they really need to change it, they should contact us so we can handle it manually.


packages/core/src/modules/attendance/__test__/attendance-service.e2e-spec.ts
```diff
@@ -3,17 +3,16 @@ import { AttendancePoolWrite, AttendanceWrite, AttendeeWrite, EventWrite, UserWr
 import crypto from "crypto"
 import { ulid } from "ulid"
 import { afterEach, beforeEach, describe, expect, it } from "vitest"
+import assert from "../../../../assert"
 import { CleanupFunction, createServiceLayerForTesting } from "../../../../vitest-integration.setup"
-import { createServiceLayer, type ServiceLayer } from "../../core"
+import { type ServiceLayer, createServiceLayer } from "../../core"
```
henrikhorluck: Can we get some import sorting into CI?

henrikskog: It would be really nice to have biome fix imports, I don't think it does any import checking atm

henrikskog:
Pretty sure the web deployment uses the production database, where all events are missing attendanceId, thus the validation flops. So CI is passing here

packages/core/src/modules/attendance/__test__/attendance-service.e2e-spec.ts
henrikskog: https://linear.app/dotkom/issue/DOT-690/make-biome-sort-and-fix-imports

PR #849, opened 2024-03-19
Auth0 updates

henrikhorluck:
- **Upgrade Auth0 provider**
- **Add a custom login background pattern**
- **Fix vengeful callback**
- **Allow logout for android**
- **Enable refresh token rotation**

Closes: #...

## Checklist

- [ ] I have written tests

## Changelog

- A bullet list of what you've done to make the life of the reviewer easier.

## How to test

A consise description of how to test the newly implemented functionality.


henrikskog (commented):
Nice! 

> [rotation_type](https://registry.terraform.io/providers/auth0/auth0/latest/docs/resources/client#rotation_type) (String) Options include rotating, non-rotating. When rotating, exchanging a refresh token will cause a new refresh token to be issued and the existing token will be invalidated. This allows for automatic detection of token reuse if the token is leaked.

Do you know if the automatic detection is set up automatically or is this something we would need to set up ourselves to be able to track leaks?

infra/auth0/main.tf
```diff
@@ -61,6 +61,10 @@ resource "auth0_branding" "branding" {
     # online-blue
     page_background = "#0D5474"
   }
+
+  universal_login {
```
henrikskog, line 65: Curious about this, what does it do

henrikhorluck: This is the easter background, I just forgot to push it

junlarsen (approved):

infra/auth0/main.tf
henrikskog: Nice! And creds for the commits, really pleasant to review

PR #855, opened 2024-04-01
Add sample data importer and prototype for frontpage

jotjern:
image

This is just a prototype, but I think having a prototype to work on and improve is better than having just a bunch of JSON data. The splash here is mostly for companies, but my thought is that we remove it when a user is signed in (we don't need to explain what Online is to people who have Onlineweb-accounts), and replace it with maybe like a single large promoted event, or something other we want to showcase (interest groups, announcements, etc...). In the fadderweeks we can also show a large promo for our splash page (which we need to make btw). 

The sample data importer needs to be fixed when events-POC is merged.


apps/web/src/components/molecules/ComingEvent/ComingEvent.tsx
```diff
@@ -14,28 +15,32 @@ interface ComingEventProps {
...
+        <Badge color="green" variant="solid" className="absolute bottom-2 left-2">
+          {props.tag}
+        </Badge>
+      </div>
+      <div className="">
+        <span>
+          <span>{/* PLACEHOLDER */}</span>
+          <div>
+            <p className="m-0">{props.title}</p>
+            <p className="m-0">{props.date}</p>
```
henrikhorluck, line 37: I think datetime-formatting should be determined here?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat

tools/migrator/src/ow4-sample-data/ow4-events.ts
```diff
@@ -0,0 +1,121 @@
...
+  rule_bundles: z.array(z.any()),
+  number_on_waitlist: z.number(),
+  number_of_seats_taken: z.number(),
+  extras: z.array(z.any()),
+})
+
+const OW4APIEventSchema = z.object({
+  absolute_url: z.string(),
+  attendance_event: OW4APIAttendanceEventSchema.nullable(),
+  company_event: z.array(z.any()), // Adjust based on the actual structure of company events
```
henrikhorluck, line 42: Feltet heter `companies` og er en liste med navnene på selskapene

apps/web/src/app/page.tsx
```diff
@@ -1,17 +1,42 @@
...
+  const session = await getServerSession()
+  const events = await serverClient.event.recommended()
 
   return (
     <div>
-      <h1>Auth:</h1>
-      <pre>{JSON.stringify(auth, null, 4)}</pre>
-      <h1>Events:</h1>
-      <pre>{JSON.stringify(events, null, 4)}</pre>
+      <CompanySplash className="hidden md:flex" />
```
henrikhorluck, line 15: Kan vi få inn at det er egen side for innlogget vs utlogget allerede nå?

tools/migrator/src/ow4-sample-data/ow4-events.ts
jotjern: Vi bruker den ikke, schema trenger ikke være bra imo når den bare er til sample data

packages/gateway-trpc/src/modules/event/event-router.ts
```diff
@@ -58,6 +58,18 @@ export const eventRouter = t.router({
       committees: results[i],
     }))
   }),
+  recommended: publicProcedure.query(async ({ ctx }) => {
+    const events = await ctx.eventService.getEvents(4)
+    const committees = events.map(async (e) => ctx.eventCommitteeService.getEventCommitteesForEvent(e.id))
```
junlarsen, line 63: N+1 Query

jotjern: True that, kokte fra all routen, men burde fikses

apps/web/src/app/page.tsx
jotjern: Det kan vi se på ja

apps/web/src/components/molecules/ComingEvent/ComingEvent.tsx
jotjern: ComingEvent trenger en del arbeid ja

junlarsen (approved):

PR #858, opened 2024-04-04
Fix query error with tRPC in Next 14 Route Handler

junlarsen:
tRPC doesn't behave well as a Route Handler, presumably because it doesn't handle OPTIONS or something properly


henrikskog (approved):

apps/web/src/utils/trpc/serverClient.ts
```diff
@@ -1,7 +1,8 @@
 "use server"
 
-import { appRouter, createCallerFactory, createContext, createUnauthorizedContext } from "@dotkomonline/gateway-trpc"
+import { appRouter, createCallerFactory, createContextInner } from "@dotkomonline/gateway-trpc"
 
 const createCaller = createCallerFactory(appRouter)
-export const getServerClient = async () => createCaller(await createContext())
-export const getUnauthorizedServerClient = async () => createCaller(await createUnauthorizedContext())
+// TODO: Add a way to get the userId from the request
+export const getServerClient = async () => createCaller(createContextInner({ auth: { userId: null } }))
```
junlarsen, line 7: Of course, this is wrong, but I don't have time to figure out what needs to be added here.

https://linear.app/dotkom/issue/DOT-719/make-getserverclient-call-createcaller-with-the-currently-requested tracks this

PR #863, opened 2024-04-08
im batman

henrikskog:
**Running in "prod" (though called staging)**

Source code for the batman lambda function. This is very wip. but it is working and running in the kontor.

I'm honestly not sure if we want this merged or not. We'll know more about the future of the kiosk in a meeting tomorrow so then I will decide if it is worth it to clean up and get it merged.


henrikhorluck (changes requested):
Liksom et minimum av README eller beskrivelse av hva dette er??

PR #866, opened 2024-04-10
use fake user data before additional user data is gathered

henrikskog:


packages/auth/src/auth-options.ts
```diff
@@ -1,39 +1,35 @@
...
 
-  interface User extends DefaultUser {
-    id: string
-    name: string
-    email: string
-    image?: string
-    givenName?: string
-    familyName?: string
+  interface User extends IdToken {
+    email: string // Without specifying a random new field, the User type is not augmented. Why? :(
```
henrikskog, line 32: Do you know if this can be handled better?

henrikskog:
This was not built upon the latest version of main 🛑 Opening new PR

PR #867, opened 2024-04-11
feature/dot 630 fix domain user model and populate with fake data 3

henrikskog:
### Create dashboard page for managing users that communicates with auth0 and local db
image
image

- See all users
- Update info about users that is not managed by auth0. I.e. not email, id, createdAt, updatedAt, email_verified

### Store application user data in app_meta and expose to applications in jwt in next-auth
- If a user is fetched from auth0 and they don't have user data like familyName etc, this data is populated with fake data. We remove this when we make the users provide more data on sign up. This way we avoid having to deal with these fields being null in the backend when they will be enforced in the future.
- I added a flow in auth0 that adds all the custom fields from app_metadata to the token post login. **This is not in terraform yet**. TODODODOD

### Other decisions that can be discussed
- Remove ulid PK in ow_user and use the auth0 id as pk instead. I called it auth0_id to be explicit about where it comes from. 
- Use app_metadata instead of user_metadata to store user data in auth0. Users can edit data in user_metadata with their access tokens, but not stuff in app_metadata. I.e. we don't want to store the users `studyYear` in user_metadata of this reason. So I just went for the easiest solution storing everything in `app_metadata` for now, as the users are not communicating directly with the api anyway, everything goes thorugh the backend.
  - Right now we also store familyName, givenName and gender in app_metadata. This may be better to store as normal claims so that they are available for i.e. vinstraff.no


tools/migrator/src/seed.ts
henrikskog, line 1: old seed data. AFAIK this is not used so I did not care to fix errors here after changing owUser to contain more fields. Thoughts?

tools/migrator/src/fixtures/user.ts
henrikskog, line 1: user data is automatically inserted intot the db when authenticating with auth0 so it does not really make sense to have these fixtures here anymore

tools/migrator/src/fixtures/personal-mark.ts
henrikskog, line 1: removed user fixtures

apps/web/src/components/views/SettingsView/components/SettingsLanding.tsx
```diff
@@ -17,6 +17,9 @@ const FormInput: React.FC<FormInputProps> = ({ title, children }) => (
 )
 
 const Landing: NextPage<{ user: User }> = ({ user }) => {
```
henrikskog, line 19: We now have access to these values here in web from http cookie!

apps/web/src/app/settings/page.tsx
```diff
@@ -1,9 +1,10 @@
 import { SettingsLanding } from "@/components/views/SettingsView/components"
+import { web as authOptions } from "@dotkomonline/auth"
 import { getServerSession } from "next-auth"
 import { redirect } from "next/navigation"
 
 const SettingsPage = async () => {
-  const session = await getServerSession()
+  const session = await getServerSession(authOptions)
```
henrikskog, line 7: Important to pass with authOptions here, this was forgotten from before

packages/core/src/modules/attendance/attendance-service.ts
```diff
@@ -121,21 +121,22 @@ export class AttendanceServiceImpl implements AttendanceService {
...
+      await this.waitlistAttendeeRepository.update(waitlistAttendee.id, { attendancePoolId: mergePool.id })
+    }
 
-    await Promise.all(updatedAttendees)
-    await Promise.all(poolsToDelete)
-    await Promise.all(updatedWaitlistAttendees)
+    for (const pool of pools) {
+      await this.attendancePoolRepository.delete(pool.id)
+    }
   }
```
henrikskog, line 141: This is flaky. Sometimes the third update of waitlist attendees fail. The buggy state with this code is in https://github.com/dotkom/monoweb/pull/new/example-buggy-promise-all. Running the tests you will see that sometimes this fails and sometimes not. Weird stuff.

apps/dashboard/src/app/(dashboard)/punishment/[id]/page.tsx
```diff
@@ -8,13 +8,13 @@ import type { FC } from "react"
...
-import { useUserGetQuery } from "../../../../modules/user/queries/use-user-get-query"
+import { useUserQuery } from "../../../../modules/user/queries"
 import { useMarkWriteForm } from "../write-form"
 import { useMarkDetailsContext } from "./provider"
 
 const UserNameCell: FC<{ userId: string }> = ({ userId }) => {
-  const { user } = useUserGetQuery(userId)
-  return <span>{user?.id || "Loading..."}</span>
+  const { data: user } = useUserQuery(userId)
+  return <span>{user?.auth0Id || "Loading..."}</span>
```
Comment (henrikhorluck) [inline apps/dashboard/src/app/(dashboard)/punishment/[id]/page.tsx:17]: E-posten er allerede unik, jeg ville egt. ikke vist auth0-id-en til noen utenfor dotkom

Comment (henrikskog) [reply apps/dashboard/src/app/(dashboard)/punishment/[id]/page.tsx:17]: Ah, I did not touch this part of the code in this PR. Looks like an N+1 query, yep.

Comment (henrikskog) [reply apps/dashboard/src/app/(dashboard)/punishment/[id]/page.tsx:17]: For the auth0 id, this does not seem like a big concern if it is shown in a dashboard for admin users? I saw some discussion about it here https://community.auth0.com/t/are-auth0-ids-senstive/71117/6. 

But I agree that the general consensus should be that you don't expose it though. This is just left over code from when we did not have other data to show maybe

Comment (henrikhorluck) [reply apps/dashboard/src/app/(dashboard)/punishment/[id]/page.tsx:17]: Jeg ville heller bare holdt det skjult siden det er en implementasjonsdetalj de ikke trenger vite om, og det er nok mer intuitivt å se e-posten (men atm kan jo den endres da forsåvidt). Hva de ikke vet har de ikke vondt av

apps/web/src/components/views/SettingsView/components/SettingsLanding.tsx
henrikskog: commented on the line above what i meant oops

tools/migrator/src/index.ts
```diff
@@ -65,7 +65,7 @@ program
             .join("\n")}`
         )
       } else {
-        logger.warn(res)
+        console.dir(res.error, { depth: null })
```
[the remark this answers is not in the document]
junlarsen: add a `winston.format.splat()` to the packages/logger/src/index.ts, and do this instead

```ts
logger.warn("error performing database migrations: %O", res)

File: packages/core/src/modules/attendance/attendance-service.ts (hunk 3 above)
Comment (junlarsen) [reply packages/core/src/modules/attendance/attendance-service.ts:141]: This is a clear race condition from the above code. I think this would just as well be solved with ```ts
await Promise.allSettled(attendees.map(...))
await Promise.allSettled(b.map(...))
await Promise.allSettled(c.map(...))
```

henrikskog: The three `Promise.all`'s are not dependant on each other though? I don't understand how the race condition occurs

tools/migrator/src/index.ts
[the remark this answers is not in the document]
henrikskog: Ah nice!

henrikskog:
I'm stackign this PR to make it easier to review 👍

packages/core/src/modules/attendance/attendance-service.ts
junlarsen: Oh they don't? Then I don't see why performing them in a loop over Promise.all matters

PR #868, opened 2024-04-12
Auth0 infra: update Auth0 GTX client scope

henrikskog:


junlarsen (commented):
do we not need the read:user_idp_tokens scope anymore?

henrikskog:
No. The client only reads and updates users. I don't think that grant should have been there from the start.

PR #869, opened 2024-04-12
User Attributes Issue: Add custom claims to id token from auth0

henrikskog:
Adds action that adds user data from the user metadata to the id token post login.

### DIFF
--- a/infra/auth0/js/setCustomIdTokenClaims.js
+++ b/infra/auth0/js/setCustomIdTokenClaims.js
@@ -5,7 +5,6 @@
  * @param {PostLoginAPI} api - Interface whose methods can be used to change the behavior of the login.
  */
 exports.onExecutePostLogin = async (event, api) => {
-  api.idToken.setCustomClaim("userId", event.user.user_id)
   api.idToken.setCustomClaim("gender", event.user.user_metadata.gender)
   api.idToken.setCustomClaim("studyYear", event.user.app_metadata.studyYear)
   api.idToken.setCustomClaim("phoneNumber", event.user.app_metadata.phoneNumber)
@@ -14,10 +13,5 @@ exports.onExecutePostLogin = async (event, api) => {
   api.idToken.setCustomClaim("givenName", event.user.app_metadata.givenName)
   api.idToken.setCustomClaim("familyName", event.user.app_metadata.familyName)
   api.idToken.setCustomClaim("name", event.user.app_metadata.name)
-  api.idToken.setCustomClaim("emailVerified", event.user.app_metadata.emailVerified)
   api.idToken.setCustomClaim("onBoarded", event.user.app_metadata.onBoarded)
-
-  api.idToken.setCustomClaim("id", event.user.user_id)
-  api.idToken.setCustomClaim("user_id", event.user.user_id)
-  api.idToken.setCustomClaim("sub", event.user.user_id)
 }
--- a/infra/auth0/js/setCustomIdTokenClaims.js
+++ b/infra/auth0/js/setCustomIdTokenClaims.js
@@ -8,6 +8,7 @@ exports.onExecutePostLogin = async (event, api) => {
   const namespace = "https://online.ntnu.no"
 
   const { study_year, ow_user_id } = event.user.app_metadata
+  const { middle_name } = event.user.user_metadata
 
   if (study_year) {
     api.idToken.setCustomClaim(`${namespace}/study_year`, study_year)
@@ -15,5 +16,9 @@ exports.onExecutePostLogin = async (event, api) => {
 
   if (ow_user_id) {
     api.idToken.setCustomClaim(`${namespace}/ow_user_id`, ow_user_id)
+  }
+
+  if (middle_name) {
+    api.idToken.setCustomClaim(`${namespace}/middle_name`, middle_name)
   }
 }
--- a/infra/auth0/js/setCustomIdTokenClaims.js
+++ b/infra/auth0/js/setCustomIdTokenClaims.js
@@ -6,11 +6,14 @@
  */
 exports.onExecutePostLogin = async (event, api) => {
   const namespace = "https://online.ntnu.no"
-  const { gender, phone, allergies } = event.user.app_metadata
-  const { study_year } = event.user.user_metadata
 
-  api.idToken.setCustomClaim(`${namespace}/gender`, gender)
-  api.idToken.setCustomClaim(`${namespace}/phone`, phone)
-  api.idToken.setCustomClaim(`${namespace}/study_year`, study_year)
-  api.idToken.setCustomClaim(`${namespace}/allergies`, allergies)
+  const { study_year, ow_user_id } = event.user.app_metadata
+
+  if (study_year) {
+    api.idToken.setCustomClaim(`${namespace}/study_year`, study_year)
+  }
+
+  if (ow_user_id) {
+    api.idToken.setCustomClaim(`${namespace}/ow_user_id`, ow_user_id)
+  }
 }
--- a/infra/auth0/js/setCustomIdTokenClaims.js
+++ b/infra/auth0/js/setCustomIdTokenClaims.js
@@ -6,17 +6,11 @@
  */
 exports.onExecutePostLogin = async (event, api) => {
   const namespace = "https://online.ntnu.no"
-  const { gender, studyYear, phoneNumber, allergies, profilePicture, givenName, familyName, name, onBoarded } =
-    event.user.app_metadata
+  const { gender, studyYear, phoneNumber, allergies, onBoarded } = event.user.app_metadata
 
-  api.idToken.setCustomClaim(`${namespace}/givenName`, givenName)
-  api.idToken.setCustomClaim(`${namespace}/familyName`, familyName)
-  api.idToken.setCustomClaim(`${namespace}/name`, name)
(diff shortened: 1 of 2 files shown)


infra/auth0/js/setCustomIdTokenClaims.js
```diff
@@ -0,0 +1,23 @@
...
+  api.idToken.setCustomClaim("profilePicture", event.user.app_metadata.profilePicture)
+  api.idToken.setCustomClaim("givenName", event.user.app_metadata.givenName)
+  api.idToken.setCustomClaim("familyName", event.user.app_metadata.familyName)
+  api.idToken.setCustomClaim("name", event.user.app_metadata.name)
+  api.idToken.setCustomClaim("emailVerified", event.user.app_metadata.emailVerified)
+  api.idToken.setCustomClaim("onBoarded", event.user.app_metadata.onBoarded)
+
+  api.idToken.setCustomClaim("id", event.user.user_id)
+  api.idToken.setCustomClaim("user_id", event.user.user_id)
+  api.idToken.setCustomClaim("sub", event.user.user_id)
```
henrikhorluck, line 22: https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

- `snake_case` ikke `camelCase` så `phone_number` ikke `phoneNumber`
- De aller fleste av disse _er_ allerede i id_tokenen?
- Hvorfor lagres alt i app_metadata?
- Hvorfor er det 4 felter til et id-felt, når vi allerede bruker sub-en?

henrikhorluck (changes requested):

infra/auth0/js/setCustomIdTokenClaims.js
```diff
@@ -0,0 +1,17 @@
...
+ */
+exports.onExecutePostLogin = async (event, api) => {
+  api.idToken.setCustomClaim("gender", event.user.user_metadata.gender)
+  api.idToken.setCustomClaim("studyYear", event.user.app_metadata.studyYear)
+  api.idToken.setCustomClaim("phoneNumber", event.user.app_metadata.phoneNumber)
+  api.idToken.setCustomClaim("allergies", event.user.app_metadata.allergies)
+  api.idToken.setCustomClaim("profilePicture", event.user.app_metadata.profilePicture)
+  api.idToken.setCustomClaim("givenName", event.user.app_metadata.givenName)
+  api.idToken.setCustomClaim("familyName", event.user.app_metadata.familyName)
+  api.idToken.setCustomClaim("name", event.user.app_metadata.name)
```
henrikhorluck, line 15: Alle disse er allerede i `id_token` fra før, (merk at navn er e-posten by default, men det er noe vi kan endre), vi kan [overskrive](https://auth0.com/docs/manage-users/user-accounts/change-user-picture) `picture` — defaulten peker bare til gravatar

henrikhorluck: Du kan se hva som er i id_token ved å logge inn på OWF go se i session storage, og parse den på f.eks. https://jwt.io

infra/auth0/js/setCustomIdTokenClaims.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Handler that will be called during the execution of a PostLogin flow.
+ *
+ * @param {Event} event - Details about the user and the context in which they are logging in.
+ * @param {PostLoginAPI} api - Interface whose methods can be used to change the behavior of the login.
+ */
+exports.onExecutePostLogin = async (event, api) => {
+  api.idToken.setCustomClaim("gender", event.user.user_metadata.gender)
```
henrikhorluck, line 8: `gender` får vi ikke lov til å sette av auth0, må f.eks. namespaces til `https://online.ntnu.no/gender` https://auth0.com/docs/secure/tokens/json-web-tokens/create-custom-claims#general-restrictions

junlarsen: I think it's probably best to prefix all of the custom claims anyways. That's pretty common practice.

Suggest we do `https://online.ntnu.no/x` for all of them

henrikskog:
I took a look at the documentation and I think we should avoid storing attributes in user metadata that can be stored as root attributes. I removed these claims. 

I'm not sure if we should store the `phone_number` claim in metadata or not. https://community.auth0.com/t/why-can-we-not-set-a-users-phone-number/48574

henrikhorluck:
> I'm not sure if we should store the `phone_number` claim in metadata or not. This forum post is from 2020 but states that the root `phone_number` attribute is reserved for `sms users`, whatever that means. https://community.auth0.com/t/why-can-we-not-set-a-users-phone-number/48574.

As far as I understood it is only allowed to be used for users that have passwordless [log in with a phone number](https://auth0.com/docs/authenticate/passwordless/authentication-methods/sms-otp). We do not have such users, so the root `phone_number` is not relevant to us (yes that is a bit weird, but w/e), and we need to store it in app/user_metadata

infra/auth0/js/setCustomIdTokenClaims.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Handler that will be called during the execution of a PostLogin flow.
+ *
+ * @param {Event} event - Details about the user and the context in which they are logging in.
+ * @param {PostLoginAPI} api - Interface whose methods can be used to change the behavior of the login.
+ */
+exports.onExecutePostLogin = async (event, api) => {
+  const namespace = "https://online.ntnu.no"
+  const { gender, studyYear, phoneNumber, allergies } = event.user.app_metadata
```
henrikhorluck, line 9: `phone` is in `user_metadata`.

Rule-of-thumb: user-metadata = user can freely change, app-metadata: system relies on it for something and has to be protected

allergies and gender should also be in user-metadata, study-year would be app-metadata. IMO study-year could just be behind a `${namespace}/membership`, which is wholly managed by the membership-service, which could then contain e.g. `study_year`, `field_of_study`, `expiry_data`, etc.

henrikskog: > Rule-of-thumb: user-metadata = user can freely change, app-metadata: system relies on it for something and has to be protected

Is there any use case for this except allowing users to update data in `user_metadata` in SPA's? (which auth0 does not recommend btw https://auth0.com/docs/secure/tokens/access-tokens/management-api-access-tokens/get-management-api-tokens-for-single-page-applications)

IMO it's easier to just store everything in `app_metadata`

henrikhorluck: I think it makes sense to split it, I do not think it causes any overhead.

henrikskog: Sure, it does not really matter and you already put phone in `user_metadata` so why not just follow their convention.

infra/auth0/js/setCustomIdTokenClaims.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Handler that will be called during the execution of a PostLogin flow.
+ *
+ * @param {Event} event - Details about the user and the context in which they are logging in.
+ * @param {PostLoginAPI} api - Interface whose methods can be used to change the behavior of the login.
+ */
+exports.onExecutePostLogin = async (event, api) => {
+  const namespace = "https://online.ntnu.no"
+  const { gender, phone, allergies } = event.user.app_metadata
+  const { study_year } = event.user.user_metadata
```
henrikhorluck, line 10: Doesn't this crash if study_year is not present?

henrikskog: I thought about it but IMO it's okay to just have it as undefined if it is not set?

henrikskog: If the attributes don't exist the values will resolve to undefined, which again will resolve to null in the JWT I assume. I'll test it. I think it makes a lot of sense that the values are null

henrikskog: This code will raise an exception if app_metadata or user_medatata is undefined, but this should never be the case

henrikskog:
I think we only need `studyYear` and `ow_user_id` in custom claims. Let's fetch the other ones if they are needed.

infra/auth0/js/setCustomIdTokenClaims.js
```diff
@@ -0,0 +1,24 @@
...
+
+  const { study_year, ow_user_id } = event.user.app_metadata
+  const { middle_name } = event.user.user_metadata
+
+  if (study_year) {
+    api.idToken.setCustomClaim(`${namespace}/study_year`, study_year)
+  }
+
+  if (ow_user_id) {
+    api.idToken.setCustomClaim(`${namespace}/ow_user_id`, ow_user_id)
```
henrikhorluck, line 18: Hvilke OW er dette ID-en til?

henrikskog: Ow5

henrikhorluck: Hvorfor har OW5 egne bruker-ID-er? bestemte vi oss ikke for at vi bare bruke Auth0-sub-en som ID?

henrikskog: For å decouple brukerene våre fra auth0. Vi snakket om det ja, men jeg og Mats ble enige om at det er nice å ikke fylle databasen med user FK's som er auth0-id'er

PR #870, opened 2024-04-12
Fix race condition when merging attendance

henrikskog:


packages/core/src/modules/attendance/attendance-service.ts
```diff
@@ -113,29 +110,26 @@ export class AttendanceServiceImpl implements AttendanceService {
...
 
-    // move all waitlist attendees to the new pool, sort by position value
-    const waitlistAttendees = await this.waitlistAttendeeRepository.getByAttendanceId(attendanceId)
-    const combinedWaitlistAttendees = waitlistAttendees.sort((a, b) => a.position - b.position)
-    const updatedWaitlistAttendees = combinedWaitlistAttendees.map((waitlistAttendee, i) =>
-      this.waitlistAttendeeRepository.update({ attendancePoolId: mergePool.id, position: i }, waitlistAttendee.id)
+    const waitlistAttendeesUnordered = await this.waitlistAttendeeRepository.getByAttendanceId(attendanceId)
+    const waitlistAttendeesOrdered = waitlistAttendeesUnordered.sort(
+      (a, b) => a.registeredAt.getTime() - b.registeredAt.getTime()
     )
```
henrikskog, line 119: Only this is related to the race condition, code above is just cleanup

junlarsen (approved):
Happy to accept as-is, but could you try the Promise.all thing I suggested in #867? Otherwise I'm at a loss for how we have a race condition

henrikskog:
I've tested this thourougly now and it works now using `promise.all`.

junlarsen (approved):

PR #871, opened 2024-04-12
Store and handle more user attributes

henrikskog:
- Stores all user attributes and not just email and studyYear
- Removes user fixtures because they are outdated and of little value now that users are stored in auth0
- Removes user dashboard page as this is recreated in next PR to avoid having 1 million changes in this PR

For how user data is stored in auth0 (some properties are stored in `app_metadata` and some in `user_metadata`, see: https://github.com/dotkom/monoweb/blob/7aced048719598e1ad93aa2137851ad249a4afe8/packages/core/src/modules/external/auth0-repository.ts#L17 

I have added a function that inserts fake data for the new fields until we gather this on signup such that we can maintain the not-null integrity of these fields thorughout the backend. See https://github.com/dotkom/monoweb/blob/7aced048719598e1ad93aa2137851ad249a4afe8/packages/auth/src/auth-options.ts#L77


packages/core/mock.ts
```diff
@@ -1,15 +1,40 @@
 import type { CompanyWrite, JobListingWrite, UserWrite } from "@dotkomonline/types"
+import type { GetUsers200ResponseOneOfInner } from "auth0"
 import { addWeeks, addYears } from "date-fns"
+import { ulid } from "ulid"
+
+export const getAuth0UserMock = (write?: Partial<UserWrite>): GetUsers200ResponseOneOfInner =>
+  ({
+    user_id: write?.auth0Id ?? "auth0|test",
+    email: write?.email ?? "fakeemail@gmai.com",
+    given_name: write?.givenName ?? "Ola",
```
madsab, line 10: Should <code>given_name</code> include the middle name as we don't have a separate field for it. This way it's possible to update the middle-name as well.
 ```suggestion
    given_name: write?.givenName ?? "Ola Mellomnavn",
```

Comment (henrikskog) [reply packages/core/mock.ts:10]: How we have it now we have three fields related to the name. 
- given_name = first name
- family_name = last name
- name = full name including possible middle names. 

We will probably just change name to be given+family every time one of them changes? 

IMO we can just let people add their middle names to their given_name if they want maybe?

Comment (henrikskog) [reply packages/core/mock.ts:10]: I didn't know middle_name was a standard claim. So in this case yeah I agree let's add middle name

Comment (junlarsen) [reply packages/core/mock.ts:10]: Well the standard claims are all optional (except for sub), so that's totally fine and compliant.

Comment (henrikskog) [reply packages/core/mock.ts:10]: Okay yeah true. Middle name is now stored as it's own field in `user_metadata`! 👍

File: packages/core/src/modules/attendance/__test__/attendance-service.e2e-spec.ts [hunk 2]
```diff
@@ -141,9 +132,9 @@ describe("attendance", () => {
 
     expect(pools).toHaveLength(1)
 
-    const fakeUser = getFakeUser({ studyYear: 1 })
+    const fakeUser = getUserMock({ studyYear: 1 })
 
-    const user = await core.userService.createUser(fakeUser)
+    const user = await core.userRepository.create(fakeUser)
```
junlarsen, line 137: any reason for this change? it seems fair to call the service here

packages/core/src/modules/core.ts
```diff
@@ -251,5 +255,6 @@ export const createServiceLayer = async ({ db }: ServerLayerOptions) => {
     interestGroupService,
     auth0Repository,
     auth0SynchronizationService,
+    userRepository,
```
junlarsen, line 258: never export repositories from the service layer

packages/core/src/modules/external/auth0-synchronization-service.ts
```diff
@@ -15,69 +17,70 @@ export interface Auth0SynchronizationService {
...
+    const updatedUser: UserWrite = {
+      ...userAuth0,
+      lastSyncedAt: new Date(),
     }
 
-    const user = await this.userService.getUserBySubject(auth0Subject)
-
-    if (user === undefined) {
-      this.logger.log("info", "User does not exist in local db, creating user", { userId: auth0Subject })
+    const userDb = await this.userRepository.getById(userAuth0.id)
```
junlarsen, line 52: This would preferably use the UserService

packages/core/src/modules/user/__test__/core.ts
```diff
@@ -0,0 +1,55 @@
...
+import { type UserService, UserServiceImpl } from "../user-service"
+
+export type ServiceLayer = Awaited<ReturnType<typeof createServiceLayerForUserTests>>
+
+export interface ServerLayerOptions {
+  db: Kysely<Database>
+  auth0MgmtClient: ManagementClient
+}
+
+export const createServiceLayerForUserTests = async ({ db, auth0MgmtClient }: ServerLayerOptions) => {
```
junlarsen, line 25: I'd move this to vitest-integration.setup.ts or something like that

junlarsen (changes requested):
Nice work, the code looks pretty nice, but I've commented a few places where you're "breaking the rules" of our architecture

packages/core/src/modules/user/__test__/core.ts
henrikskog: This is meant to be the specific setup of the service layer for the user tests. I put it here to avoid polluting the test file. I'm not sure if this can be generalised to other tests suites but maybe? I can check out if it's worth doing.

henrikskog: I think it's easier to just create the service layer with mock injections in each test because each test will need to mock out different external sdks/repositories/services. I moved it into the test file. Leaving it like this for this PR

packages/core/src/modules/core.ts
```diff
@@ -249,7 +248,6 @@ export const createServiceLayer = async ({ db }: ServerLayerOptions) => {
     attendeeService,
     interestGroupRepository,
```
junlarsen, line 249: unrelated but this should probably not be here (cc @Terbau)

packages/core/src/modules/external/auth0-repository.ts
```diff
@@ -1,38 +1,83 @@
...
+
+    if (user.status !== 200) {
+      this.logger.error("Received non 200 status code when trying to get user from auth0", { status: user.status })
+      throw new GetUserServerError(user.statusText)
+    }
+
+    return this.mapToUser(user.data)
+  }
+
+  private mapToUser(user: GetUsers200ResponseOneOfInner): User {
```
junlarsen, line 58: this uses no dependencies from the class, move to top level scope

packages/core/src/modules/user/auth0-synchronization-service.ts
```diff
@@ -0,0 +1,104 @@
...
+import type { User, UserWrite } from "@dotkomonline/types"
+import { addDays } from "date-fns"
+import { Auth0UserNotFoundError } from "../external/auth0-errors"
+import type { Auth0Repository } from "../external/auth0-repository"
+import type { UserService } from "./user-service"
+
+export interface Auth0SynchronizationService {
+  update(payload: UserWrite): Promise<User>
+  handleUserSync(sub: string, now: Date): Promise<User>
+  handlePopulateUserWithFakeData(auth0Id: string, email?: string | null): Promise<void>
```
junlarsen, line 11: Small nit, but I don't think prefixing actions with `handle` adds any meaning to the function name. Perhaps just do `synchronizeUser()` and `populateUserWithFakeData()`.

Can you also add a doc comment on `populateUserWithFakeData()` on why exactly we have a method involving fake data? (Just so others that don't know the ins and out of the auth0 onboarding can understand?)

junlarsen (changes requested):

junlarsen (approved):

PR #878, opened 2024-04-14
Turn off retry and refetch on window focus during development

henrikskog:
I think having request retries and window refetching makes little sense during development.

### DIFF
--- a/apps/dashboard/src/app/QueryProvider.tsx
+++ b/apps/dashboard/src/app/QueryProvider.tsx
@@ -10,7 +10,7 @@ export const QueryProvider = ({ children }: PropsWithChildren) => {
       new QueryClient({
         defaultOptions: {
           queries: {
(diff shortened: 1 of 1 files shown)


junlarsen (commented):
Hmm why not? I find it pretty nice to have

henrikskog:
Well, In prod, a user may get unlucky and have their request get dropped, so there it makes a lot of sense. imo if a request fails in development, 99% of the time it won't help to retry the same request. And when it retries, you get three errors in the terminal instead of one, which can be a bit annoying when the errors are big and harder to reason about especially for beginners, which probably don't know why the retrying would be happening. 

I feel the same about windows refocus, but to a lesser degree though. It's very nice in prod, but during development I don't really see the use and I think it is just annoying when it pollutes the console with trpc requests.

junlarsen (approved):

PR #879, opened 2024-04-15
User Attributes Issue 4: Add custom user claims to id token 

henrikskog:


packages/auth/src/auth-options.ts
```diff
@@ -1,39 +1,57 @@
...
-  given_name: string
-  family_name: string
-  nickname: string
-  name: string
-  picture: string
-  gender: string
-  updated_at: string
-  email: string
-  email_verified: boolean
+// https://www.iana.org/assignments/jwt/jwt.xhtml
```
junlarsen, line 5: Not that important but I prefer we use the ID Token spec link https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

henrikskog: 👍🏻

PR #884, opened 2024-04-24
Fix fucked migrations

Terbau:
Fixes fucked things


junlarsen (commented):
:fire: :skull:

junlarsen (approved):
:fire: :skull:

PR #885, opened 2024-04-25
Improve logging package config

henrikskog:
Previously it was not possible to log out objects using the winston logger. This adds support for this using interpolation in the message. It also adds some padding to the `info` part such that messages start from the same column.

Example console output
```ts
  const logger = getLogger("service a")

  logger.error("error message")
  logger.warn("warn message")
  logger.info("info message")
  logger.http("http message")
  logger.verbose("verbose message")
  logger.debug("debug message")
  logger.silly("silly message")

  console.log("")

  logger.info("test info message from service a")
  logger.info("message with metadata2 %o", { message: "John Doe", id: 123 })

  const error = new Error("Something failed!")
  logger.error("hello world %O", error)
  ```
New
image

Old (set to only showing info and more serious):
image

### DIFF
--- a/packages/core/src/modules/external/auth0-synchronization-service.ts
+++ b/packages/core/src/modules/external/auth0-synchronization-service.ts
@@ -27,7 +27,6 @@ export class Auth0SynchronizationServiceImpl implements Auth0SynchronizationServ
   ) {}
 
   private async synchronizeUser(auth0Subject: string) {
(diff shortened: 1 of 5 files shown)


junlarsen (commented):
Nice, but we can already do object interpolation with `%O`, for example

logger.info("this is an object: %O", { name: 'jeff pesos' })

henrikskog:
Now we can use object interpolation @junlarsen 👍

junlarsen (approved):

PR #894, opened 2024-04-29
Rename migrations

jotjern:
Vet ikke hvordan vi endte opp her, men ser ikke noen annen utvei enn å nuke databasen her 💀


henrikskog:
Har det noe å si at tallene er feil

henrikskog (approved):

junlarsen:
technically it doesn't matter, but that's just a source of confusion

junlarsen (commented):
We should probably squash these sometime soon too

PR #901, opened 2024-04-30
Update to Node.js 20.x

junlarsen:
Node 18 is in maintenance mode, while 20 is in active development. We should use 20.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
       - name: Setup Node.js environment
         uses: actions/setup-node@v3
         with:
-          node-version: 20.12.2
+          node-version: 20.12.7
           cache: 'pnpm'
  
       - name: Install dependencies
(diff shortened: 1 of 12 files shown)


.github/workflows/ci.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       - name: Setup Node.js environment
         uses: actions/setup-node@v3
         with:
-          node-version: 18.18.0
+          node-version: 20.12.2
```
henrikskog, line 59: Why not 20.12.7?

package.json
```diff
@@ -35,16 +35,13 @@
     "apps/*"
   ],
   "engines": {
-    "node": ">=18.18.0",
-    "pnpm": ">=8.7.6"
+    "node": ">=20.12.2",
```
henrikskog, line 38: Why kot 20.12.7?

henrikskog (approved):
Nice! Got two questions, but looks good!

.github/workflows/ci.yml
junlarsen: probably typod lol

junlarsen:
Vercel doesn't support pnpm 9 yet

.github/workflows/ci.yml
junlarsen: node 20.12.7 isn't a thing

PR #893, opened 2024-04-29
upgrade lock file version from 6 to 9

henrikskog:
Pnpm 9 was released two weeks ago [v9.0.0](https://github.com/pnpm/pnpm/releases/tag/v9.0.0), with a new lock format

PR #903, opened 2024-05-01
Revert "Upgrade pnpm lock file version from 6 to 9 (#893)"

junlarsen:
This patch reverts three commits after the lockfile upgrade to 9.

The reason we have to revert, is because Vercel doesn't have support for PNPM 9 yet.


junlarsen:
Superseded by #904

PR #918, opened 2024-05-02
DOT-770 Image and file uploading

henrikskog:
https://github.com/dotkom/monoweb/assets/55615149/32da354a-1d22-4c3b-90c3-c54790681a7e

Unstyled

Flexible and easy to add to other parts of the dashboard.

It's possible to set a fixed aspect ratio for the crop so we can help the users upload the right aspect ratio images

You have to run `migrate` locally to run this locally as I've added a migration


henrikskog:
https://linear.app/dotkom/issue/DOT-770/image-uploading-in-dashboard

junlarsen:
Use https://mantine.dev/core/file-input/?

henrikskog:
My thinking was that this could be reused in web so I've been trying to avoid spesific dashboard dependencies such as mantine, to make it easier to port over. IMO the difference in user expirience is very small between a normal html file input and mantine version also. What do you think?

henrikskog:
It does not make sense to review this anymore cause I made a lot of changes to the code here in https://github.com/dotkom/monoweb/pull/924

PR #921, opened 2024-05-08
Fix user syncing issue

henrikskog:
The syncing service was using the ow user id from user in auth0 to decide if user exists in local db or not. However, stuff might happen such that a local user has been created but id is not updated in auth0. I'm not sure how this can happen but it just happened to me locally. Using auth0 id to check if user exists locally is safer because they will always be the same.

Also fixes some logging statements that were not using string interpolation.

### DIFF
--- a/packages/core/src/modules/user/user-repository.ts
+++ b/packages/core/src/modules/user/user-repository.ts
@@ -7,7 +7,6 @@ export const mapToUser = (payload: Selectable<Database["owUser"]>): User => User
 
 export interface UserRepository {
   getById(id: UserId): Promise<User | null>
-  getByEmail(email: string): Promise<User | null>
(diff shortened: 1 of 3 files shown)


henrikhorluck:
Do we not store the auth0 sub?

henrikskog:
Ah yes we do, good suggestion! That's better.

packages/core/src/modules/user/user-repository.ts
```diff
@@ -16,6 +16,7 @@ export interface UserRepository {
 
 export class UserRepositoryImpl implements UserRepository {
   constructor(private readonly db: Kysely<Database>) {}
+
```
henrikhorluck, line 19:
```suggestion
```

PR #925, opened 2024-05-14
Add event metadata

brage-andreas:
This PR adds:
* Showing which businesses and committees organize/host the event


junlarsen:
> Note that this PR would need to be extended with a PR adding and exposing the event author's name.

Do we want that, or is it needed? I'm not sure if showing that Brage made the event adds any value to the users?

henrikhorluck:
> > Note that this PR would need to be extended with a PR adding and exposing the event author's name.
> 
> 
> 
> Do we want that, or is it needed? I'm not sure if showing that Brage made the event adds any value to the users?

Instead of author it should be the person(s) organizing the event, som they could be contacted directly.

apps/web/src/app/events/utils.ts
junlarsen, line 1: I'm not too sure about all of the manual date calculations. I'd assume Intl.DateTimeFormat could do most of the work, and if not, I'd suggest date-fns

henrikskog:
I'm also a bit skeptical to if this is a feature we actually want. Nice design though 👍 

> Instead of author it should be the person(s) organizing the event, so they could be contacted directly.

I think this would be a really useful!

apps/web/src/app/events/components/EventInfoBox.tsx
```diff
@@ -18,11 +19,23 @@ const mapToImageAndName = (item: Committee | Company) => (
...
   return (
-    <div className="mr-10 w-full flex flex-col gap-8 md:w-[60%]">
-      <div className="flex flex-row gap-8">{[...committeeList, ...companyList]}</div>
-      <p className="bg-slate-2 p-5 text-[18px] rounded-2xl">{event.description}</p>
-    </div>
+    <section className="mr-10 w-full flex flex-col space-y-8 md:w-[60%]">
+      <section className="flex flex-row space-x-8">
+        <div className="flex flex-row space-x-4">{[...committeeList, ...companyList]}</div>
+        <div className="flex flex-row space-x-4 text-slate-9 items-center">
+          <span>Lagt ut {formatDate(event.createdAt)}</span>
```
henrikhorluck, line 32: In OW4 createdAt and publication time are two separate things, and I don't think we actually store the publication time? (Which would be the last time `is_public` was set to true. 

Aka: we supported hidden events, no need to leak creation-timestmp

PR #926, opened 2024-05-19
More flexible cursor pagination

henrikskog:
This is a suggestion for a more flexible approach to cursorbased pagination. Feedback is very welcome!

The helper function `singleColPaginatedQuery` is only for when the cursor consists of one column name.

See `packages/core/src/modules/committee/committee-repository.ts` and `packages/core/src/modules/interest-group/interest-group-repository.ts` for example usage.

If you want custom sorting you have to implement the pagination logic manually for that use case


henrikskog:
Deprecation of old method here #927

packages/core/src/modules/committee/__test__/committee-service.e2e-spec.ts
henrikskog, line 1: made this mostly for testing the pagination, I could move the pagination testing to a seperate file

packages/core/src/utils/cursor-pagination/common-cursor-utils.ts
```diff
@@ -0,0 +1,20 @@
...
+  return base64Decode(cursor)
+}
+
+type RecordWithCreatedAt = { createdAt: string }
+export function buildCreatedAtCursor<O extends RecordWithCreatedAt>(record: O): Cursor {
+  return base64Encode(record.createdAt)
+}
+export function decodeCreatedAtCursor(cursor: Cursor): string {
+  return base64Decode(cursor)
+}
```
henrikskog, line 20: these do honestly not have much value, better to inline in the `singleColPaginatedQuery` call

packages/db/src/index.ts
```diff
@@ -19,6 +19,11 @@ export const createKysely = (env: Environment) =>
       }),
     }),
     plugins: [new CamelCasePlugin()],
+    // Uncomment this to log all queries, useful for debugging
```
henrikhorluck, line 22: Make this available as an env-variable? e.g. `SQL_LOGGING=1`

packages/core/src/utils/cursor-pagination/helpers.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Add cursor based pagination to a query based on ordering by a single column, e.g. created_at.
+ *
+ * Cursors are opaque strings that represent the next record that will be fetched in the ordered set of records based on the given column.
+ * A cursor value of `null` will return no records, and is returned when there are no extra records to fetch.
+ * A cursor value of `undefined` will return the first `take` records.
+ *
+ * For more complex queries that require ordering by multiple columns, manual implementation
+ * is needed.
+ */
+export async function singleColPaginatedQuery<DB, TB extends keyof DB, T>(
```
henrikhorluck, line 39: Can't we just support `columns = ["id", "X"]`, and paginate accordingly?

```suggestion
export async function PaginatedQuery<DB, TB extends keyof DB, T>(
```

packages/core/src/utils/cursor-pagination/common-cursor-utils.ts
henrikhorluck, line 1: utils/common/utils, could be more direct

packages/core/src/modules/interest-group/interest-group-repository.ts
```diff
@@ -28,9 +30,21 @@ export class InterestGroupRepositoryImpl implements InterestGroupRepository {
...
-  async getAll(pageable: Pageable): Promise<Collection<InterestGroup>> {
-    const res = paginatedQuery(this.db.selectFrom("interestGroup").selectAll(), pageable, mapToInterestGroup)
-    return res
+  async getAll(pageable: Pageable) {
+    const query = this.db.selectFrom("interestGroup").selectAll()
+
+    const result = await singleColPaginatedQuery(query, {
+      pageable,
+      decodeCursor: decodeUlidIdCursor,
+      buildCursor: buildUlidIdCursor,
```
henrikhorluck, line 39: We can just derive these from specifying which columns to paginate on, no?

packages/core/src/utils/cursor-pagination/helpers.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  // biome-ignore lint/style/noNonNullAssertion: We know there is a record at -1 from the check above.
+  return options.buildCursor(records.at(-1)!)
+}
+
+type PaginatedQueryOptions<T> = {
+  pageable: Pageable
+  decodeCursor: (cursor: Cursor) => string
+  buildCursor: (record: T) => Cursor
```
junlarsen, line 24: Suggest we call it encodeCursor and decodeCursor

packages/core/src/utils/cursor-pagination/common-cursor-utils.ts
junlarsen, line 1: I think it's a lot simpler if we move cursor-pagination/*.ts to cursor.ts instead. I don't think this constitutes a new folder

packages/core/src/utils/cursor-pagination/helpers.ts
```diff
@@ -0,0 +1,82 @@
...
+ * A cursor value of `null` will return no records, and is returned when there are no extra records to fetch.
+ * A cursor value of `undefined` will return the first `take` records.
+ *
+ * For more complex queries that require ordering by multiple columns, manual implementation
+ * is needed.
+ */
+export async function singleColPaginatedQuery<DB, TB extends keyof DB, T>(
+  query: SelectQueryBuilder<DB, TB, T>,
+  options: PaginatedQueryOptions<T>
+) {
```
junlarsen, line 42: Most cases will use two columns; created_at DESC + id DESC

packages/db/src/index.ts
```diff
@@ -19,6 +19,11 @@ export const createKysely = (env: Environment) =>
...
+    // log(event) {
+    //   console.log(event.query.sql)
+    //   console.log(event.query.parameters)
+    // }
```
junlarsen, line 26: Put this behind an environment variable maybe?

packages/core/src/utils/cursor-pagination/helpers.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  // biome-ignore lint/style/noNonNullAssertion: We know there is a record at -1 from the check above.
+  return options.buildCursor(records.at(-1)!)
+}
+
+type PaginatedQueryOptions<T> = {
+  pageable: Pageable
+  decodeCursor: (cursor: Cursor) => string
+  buildCursor: (record: T) => Cursor
+  column: string
```
junlarsen, line 25: You can make it all super safe by taking 
```ts
type PaginatiedQueryOptions<T, DB, TB extends keyof DB, C extends keyof TB = keyof TB> = {
  column: C
}

**Review by junlarsen — CHANGES_REQUESTED** [review]
Nice work. I think in most cases we want to sort on created_at DESC and id DESC to give a strict ordering when two records have the same time stamp. Thoughts?

File: packages/core/src/utils/cursor-pagination/helpers.ts [hunk 9]
```diff
@@ -0,0 +1,82 @@
...
+
+  // Take N+1 to determine if there is a record behind `take`.
+  const pageWithNextLength = take + 1
+  let pagedQuery = query
+
+  pagedQuery = query.limit(pageWithNextLength)
+  pagedQuery = pagedQuery.orderBy(sql.raw(`${options.column} ${order}`))
+  if (cursor !== undefined) {
+    const decodedCursor = decodeCursor(cursor)
+    pagedQuery = pagedQuery.where(sql.raw(`${options.column} <= '${decodedCursor}'`))
```
henrikhorluck, line 64: This seems vulnerable to SQL-injections

packages/core/src/modules/interest-group/interest-group-repository.ts
```diff
@@ -28,9 +30,21 @@ export class InterestGroupRepositoryImpl implements InterestGroupRepository {
...
+      column: "id",
+      order: "desc",
```
henrikhorluck, line 41: IMO DRF's API for ordering is quite nice, and I think we could have something similar https://www.django-rest-framework.org/api-guide/filtering/#specifying-which-fields-may-be-ordered-against

henrikhorluck (changes requested):

packages/core/src/utils/cursor-pagination/helpers.ts
henrikskog: Then what will happen to the `where` clause? Can you make this generic to multiple columns?

packages/core/src/utils/cursor-pagination/helpers.ts
henrikskog: Then what will happen to the where clause? Can you make this generic to multiple columns?

packages/core/src/utils/cursor-pagination/helpers.ts
henrikskog: 🦺 Good job, thanks

packages/db/src/index.ts
henrikskog: Good suggestion but IMO it's just as easy to just uncomment this in the very few cases that you need it?

packages/db/src/index.ts
henrikskog: See Hørluck commnt

packages/core/src/utils/cursor-pagination/helpers.ts
henrikhorluck: I think you can just use `where (columnA, columnB) < (X, Y)` https://use-the-index-luke.com/sql/partial-results/fetch-next-page

packages/db/src/index.ts
henrikhorluck: It is _much_ more practical to use configuration to configure something https://12factor.net/config

packages/core/src/utils/cursor-pagination/helpers.ts
henrikhorluck: See my comment, I have suggestion

packages/core/src/utils/cursor-pagination/common-cursor-utils.ts
henrikhorluck: Duplicate of https://github.com/dotkom/monoweb/pull/926#discussion_r1606513424, great minds think alike

packages/core/src/index.ts
```diff
@@ -1,5 +1,6 @@
 export * from "./error"
```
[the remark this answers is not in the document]
junlarsen: Barreling doesn't really matter on the server

packages/core/src/utils/cursor-pagination/types.ts
```diff
@@ -0,0 +1,20 @@
+import { z } from "zod"
+
+export interface Collection<T> {
```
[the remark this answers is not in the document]
junlarsen: lgtm

packages/db/src/index.ts
junlarsen: agreed

packages/core/src/utils/cursor-pagination/helpers.ts
junlarsen: Should be very much possible, just need to rearrange the sort orders too

packages/core/src/modules/interest-group/interest-group-repository.ts
junlarsen: Maybe, but I think that adds significant type complexity

packages/core/src/utils/cursor-pagination/helpers.ts
junlarsen: Fixed

packages/core/src/utils/cursor-pagination/helpers.ts
```diff
@@ -55,13 +57,11 @@ export async function singleColPaginatedQuery<DB, TB extends keyof DB, T>(
...
-  let pagedQuery = query
-
-  pagedQuery = query.limit(pageWithNextLength)
-  pagedQuery = pagedQuery.orderBy(sql.raw(`${options.column} ${order}`))
+  let pagedQuery = query.limit(pageWithNextLength)
+  pagedQuery = pagedQuery.orderBy(options.column, order)
   if (cursor !== undefined) {
     const decodedCursor = decodeCursor(cursor)
-    pagedQuery = pagedQuery.where(sql.raw(`${options.column} <= '${decodedCursor}'`))
+    pagedQuery = pagedQuery.where(options.column, order === "asc" ? ">" : "<", decodedCursor)
```
henrikskog: Right now the `getNextCursor` function assigns the next cursor to the take+1 record. So it should be >= and <= right? Or else we skip the take+1 record?

packages/core/src/modules/interest-group/interest-group-repository.ts
henrikhorluck: I think losing typing for a better API is a worthwhile tradeoff

packages/core/src/utils/cursor-pagination/helpers.ts
junlarsen: Yes, fixed it in https://github.com/dotkom/monoweb/pull/926/commits/a9bc2e0ddd4fa1814a13ec138cd0fe5c1e5b8690

packages/core/src/modules/interest-group/interest-group-repository.ts
junlarsen: If we drop ulids and have id+created_at for everything then we can default it to that ,and have an option for overriding

packages/core/src/utils/cursor-pagination/helpers.ts
```diff
@@ -0,0 +1,82 @@
...
+type PaginatedQueryOptions<T> = {
+  pageable: Pageable
+  decodeCursor: (cursor: Cursor) => string
+  buildCursor: (record: T) => Cursor
+  column: string
+  order: "asc" | "desc"
+}
+
+/**
+ * Add cursor based pagination to a query based on ordering by a single column, e.g. created_at.
```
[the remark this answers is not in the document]
junlarsen: Agree, slack link good

packages/core/src/modules/interest-group/interest-group-repository.ts
henrikskog: sounds good

packages/db/src/index.ts
junlarsen: just use process.env don't make them required

henrikskog: Yeah true

henrikskog: fixed

packages/core/src/modules/interest-group/interest-group-repository.ts
henrikskog: What about just using the order of cols given to `column` when that option supports multiple columns and not just one like now

junlarsen: fair, but how do you determine if the default sort is ASC or DESC?

henrikskog: Passing order like this

packages/core/src/utils/cursor-pagination/common-cursor-utils.ts
henrikskog: Never mind, I think these are nice

packages/core/src/utils/cursor-pagination/helpers.ts
henrikskog: Seeing that it takes a record and not a cursor I didn't like encodeCursor. thoughts?

packages/core/src/utils/cursor-pagination/types.ts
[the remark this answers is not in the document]
henrikskog: :thumbsup:

packages/core/src/utils/cursor-pagination/common-cursor-utils.ts
henrikskog: yep agreed

henrikskog: fixed

packages/core/src/modules/interest-group/interest-group-repository.ts
henrikskog: I implemented sorting by id by default now. Thoughts?

packages/core/src/utils/cursor-pagination/helpers.ts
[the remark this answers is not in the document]
henrikskog: Taking a look at this now!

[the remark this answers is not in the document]
henrikskog: Created a md doc for this now

packages/core/src/modules/interest-group/interest-group-repository.ts
henrikskog: I just realized that this implementation just recurses infinitely. On it.

henrikskog: There, now it works. Had to do some funky not null assertions though

packages/core/src/utils/cursor-pagination/helpers.ts
henrikskog: Alright, I'll take a look at it

packages/core/src/modules/interest-group/interest-group-repository.ts
henrikhorluck: I think the pagination should be generic and support a list of columns, it seems rather easy to do. Whether you make the API be `["X", { columnName: "Y", ordering: ASCENDING/DESCENDING }]`, or `["X", "-Y"]` I'm somewhat indifferent about

The cursor can just be base64 of e.g. a JSON-array like `["columnAValue", "columnBValue"]`

packages/core/src/utils/cursor.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  // biome-ignore lint/style/noNonNullAssertion: We know there is a record at -1 from the check above.
+  return options.buildCursor(records.at(-1)!)
+}
+
+type PaginatedQueryOptions<DB, TB extends keyof DB, C extends AnyColumn<DB, TB> | AnyColumnWithTable<DB, TB>, O> = {
+  pageable: Pageable
+  order: "asc" | "desc"
+  decodeCursor?: (cursor: Cursor) => OperandValueExpression<DB, TB, C>
+  buildCursor?: (record: O) => Cursor
```
henrikhorluck, line 44: Should not be necessary, as far as I can tell cursor-encode/decode is only dependent on which columns you have

packages/core/src/utils/cursor.ts
```diff
@@ -0,0 +1,133 @@
...
+function getNextCursor<T>(records: T[], options: GetNextCursorOptions<T>): Cursor | null {
+  // We fetched take+1 records to determine if there is a record behind `take`.
+  // If we got exactly `take+1` records, there is a next page.
+  const hasNextPage = records.length === options.pageable.take + 1
+  if (!hasNextPage) {
+    return null
+  }
+
+  // biome-ignore lint/style/noNonNullAssertion: We know there is a record at -1 from the check above.
+  return options.buildCursor(records.at(-1)!)
```
henrikhorluck, line 37: You can branch on `.at(take+1)` to avoid the noNonNullAssertions

henrikhorluck (changes requested):

packages/core/src/utils/cursor.ts
henrikskog: Good suggestion!

packages/core/src/modules/interest-group/interest-group-repository.ts
junlarsen: Yep, make cursor a base64 of some json object or array

packages/core/src/utils/cursor.ts
henrikskog: Could you explain the use case of having individual sorting on each column? Are they not the same sorting order in 99% of the time?

henrikskog: Hmm I guess if you have a table where you can sort on each column individually then it's useful.

henrikhorluck: That functionality is not what I'm describing. Though it is used quite a few places in e.g. OW4's admin-panel, mostly useful when working directly towards a table.

What I'm talking about here is rather that instead of the users of this API having to specify how to encode/decode the cursor, that should be a part the implementation detail of our CursorPagination-package/module/whatever. That would make the call-locations more focused on intent, instead of technical implementation-details (I think you would call that a leaky abstraction)

This does reduce what you can implement on top of this, but I think it is better to extend the functionality of CursorPagination rather than making call-locations more complex, and solves the general problem rather than hacking stuff together with tight coupling to e.g. our primary key-usage

henrikskog: Just fixed default cursor-encode/decode for multiple columns.

henrikskog: With the current api you cannot order on individual columns when there are multiple columns spesified, as this makes things quite a bit more complex. What do you think about this @junlarsen?

packages/core/src/utils/cursor.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  const columnSql = sql`(${sql.raw(options.columns.map((col) => `"${col}"`).join(","))})`
+  pagedQuery = pagedQuery.orderBy(columnSql, order)
+
+  if (cursor !== undefined) {
+    // biome-ignore lint/style/noNonNullAssertion: see id default vals logic above
+    const decodedCursor = decodeCursor!(cursor)
+    // pagedQuery = pagedQuery.where(options.column[0], order === "asc" ? ">=" : "<=", decodedCursor[0])
+    const _columns = sql`(${sql.raw(options.columns.map((col) => `"${col}"`).join(","))})`
+    const _values = sql`(${sql.join(decodedCursor)})`
```
henrikhorluck, line 83:
```suggestion
    const _values = sql.join(decodedCursor)
```
Does this not work? (this also needs some cleanup)

packages/core/src/utils/cursor.e2e-spec.ts
```diff
@@ -0,0 +1,114 @@
...
+const createServiceLayer = async ({ db }: ServerLayerOptions) => {
+  const committeRepository: CommitteeRepository = new CommitteeRepositoryImpl(db)
+  const committeeService: CommitteeService = new CommitteeServiceImpl(committeRepository)
+
+  return {
+    committeeService,
+  }
+}
+
+describe("cursor pagination", () => {
```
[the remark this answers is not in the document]
henrikskog: Nope, they are not ran in ci. https://linear.app/dotkom/issue/DOT-783/investigate-running-integration-tests-in-ci

Documented running the tests here: https://github.com/dotkom/monoweb/pull/929.

packages/core/src/utils/cursor.ts
henrikskog: Nope, that does not work. See https://old.kyse.link/?p=s&i=4OIEEWLXFhzj4T3b4n0I

junlarsen: You can use an expression builder (but you unfortunately cannot pass a rest to either eb.tuple or eb.tupleRef due to the types

```ts
await db
  .selectFrom('person')
  .select(['id', 'first_name'])
  .where(({ eb }) => eb(eb.refTuple(columns[0], columns[1]), '>=', eb.tuple(sql.val(values[0]), sql.val(values[1]))))
  .execute()
```

```sql
SELECT
  "id",
  "first_name"
FROM
  "person"
WHERE
  ("created_at", "id") >= ($1, $2)

-- Parameters
-- [1] 123123
-- [2] 123123
```

PR #927, opened 2024-05-19
refactor/deprecate old cursors

henrikskog:
- **Opaque cursor pagination**
- **mapping**
- **use at array helper method that supports negative indexing**
- **fixups**
- **comment**
- **deprecate old pagination function**


henrikhorluck (changes requested):
We have monorepo we can migrate it all at once, no need for deprecations.

I'd rather just migrate them all to the new scheme, isn't the operations for API-non-breaking?

henrikhorluck:
If the diff is only 

```diff
-    return await paginatedQuery(this.db.selectFrom("committee").selectAll(), pageable, mapToCommittee)
+    const query = this.db.selectFrom("committee").selectAll()
+
+    const result = await singleColPaginatedQuery(query, {
+      pageable,
+      column: "id",
+      order: "desc",
+    })
+
+    return {
+      next: result.next,
+      data: result.data.map(mapToCommittee),
+    }
```

That is doable as a regex-replace, you could also try using a Vim-macro

henrikskog:
It's not. The new scheme takes one parameter `Pageable` while the old `orderedQuery` solution use two parameters `take` and `cursor`. And many methods lack the cursor param.

junlarsen (changes requested):
I'd say we just migrate everything at once

I can look at it when I have time. It's a lot better to just move everything at once

PR #930, opened 2024-06-17
Separate tRPC from Next.js app

junlarsen:
Reapply of #835 without refresh logic

# Changes

The tRPC backend is now fully separated from the Next.js apps. This means that any first-party Next.js app can simply use `createProxyRoute` which acts as a proxy in the BFF pattern.

Refresh of access tokens not implemented.

Endpoint will be available on rpc.online.ntnu.no once infra is completed. Runs on :4444 locally


packages/gateway-trpc/src/crypto.ts
```diff
@@ -0,0 +1,26 @@
+import { createRemoteJWKSet, jwtVerify } from "jose"
+
+export class JwtService {
+  private jwks: ReturnType<typeof createRemoteJWKSet> | null = null
```
henrikhorluck, line 4:
```suggestion
  private jwks: ReturnType<typeof createRemoteJWKSet> | null = null
```
There has to be a type to use here without doing TS-magic?

packages/gateway-trpc/src/crypto.ts
```diff
@@ -0,0 +1,26 @@
...
+  public constructor(
+    private readonly issuer: string,
+    private readonly audiences: string[]
+  ) {
+    const issuerWithoutTrailingSlash = issuer.replace(/\/$/, "")
+    this.jwksUrl = new URL(`${issuerWithoutTrailingSlash}/.well-known/jwks.json`)
+  }
+
+  public async verify(accessToken: string) {
+    this.jwks ??= createRemoteJWKSet(this.jwksUrl)
```
henrikhorluck, line 16: I think that key might rotate, and has a cache-control-setting limiting how long we should cache it

if this is specifically from `https://auth.online.ntnu.no/.well-known/jwks.json` it has `Cache-Control` `public, max-age=15, stale-while-revalidate=15, stale-if-error=86400`, so we can cache it for 30s at a time

apps/rpc/src/index.ts
```diff
@@ -0,0 +1,51 @@
...
+server.register(fastifyTRPCPlugin, {
+  prefix: "/api/trpc",
+  trpcOptions: {
+    router: appRouter,
+    createContext: createFastifyContext,
+    onError({ path, error }) {
+      // report to error monitoring
+      console.error(`Error in tRPC handler on path '${path}':`, error)
+    },
+  } satisfies FastifyTRPCPluginOptions<AppRouter>["trpcOptions"],
```
henrikhorluck, line 48: Why is the `satisfies` needed? Isn't this already annotated from fastify's side?

henrikhorluck (changes requested):
Can you remind me again _why_ we need to proxy through next at all?

This needs a mermaid-diagram in docs.

I'd prefer to have a web-server like nginx/traefik/caddy as the our main entrypoint/gateway/ingress (behind Cloudflare) thing in our stack, makes it much easer to set up reverse-proxies, not sure how doable that is with Next.

packages/gateway-trpc/src/crypto.ts
junlarsen: Right can look into that, I'm used to cognito which doesn't rotate the jwks at all :upside_down_face:

apps/rpc/src/index.ts
junlarsen: For some reason TS just leaves onError object as undefined. Alternative is doing `Parameters<OnErrorFunction<AppRouter, FastifyRequest>>[0]`

image

packages/gateway-trpc/src/crypto.ts
junlarsen: Fixed in 44d5953c

henrikhorluck (approved):

junlarsen:
Will merge when infra is ready otherwise main breaks in "prod"

PR #934, opened 2024-08-11
Unify date formatting functions in a shared package

brage-andreas:
This closes DOT-788.

This PR adds:
- New package "utils"
- New functions `formatDate` and `formatRemainingTime` based off existing Intl and date-fns functions
- Changes all "local" date formatters (I found) to use this one

I'm open to feedback!


packages/utils/src/formatDate.ts
```diff
@@ -0,0 +1,70 @@
...
+} as const
+
+type SupportedLocales = keyof typeof dateFNSLocaleFromIntlLocale
+
+/**
+ * Formats a date into a string representation. If the date is within the date threshold, the date will be formatted as a relative date.
+ *
+ * @param date - The date to format.
+ * @param options - Optional formatting options.
+ * @param options.forceAbsolute - If set to true, the date will always be formatted using the `formatRemainingTime` function.
```
henrikhorluck, line 19: Is the description wrong here? `forceAbsolute` makes me think it should _never_ be relative, instead of _always_ relative? Also I think a better naming would be `neverRelative`

apps/web/src/app/events/components/TimeLocationBox/TimeLocationBox.tsx
henrikhorluck, line 1: I disagree with the changes in this file.

We should just use the output of the formatting _as is_, possibly put into a `<time>`, but we should certainly _not_ split the output based on spaces/commas and modify the capitalization, the whole point of using a formatting module is to leave the formatting to that, not duplicate that concern here.

If there you think it might end up as the start of a sentence you can use css:

```css
time {
  display: inline-block;
  &::first-letter {
    text-transform: capitalize;
  }
}
```

If you need to have it split in multiple differnt elements for styling purposes, that should rather be multiple individual calls to the formatting with arguments to only get the field you want

packages/utils/src/formatDate.ts
```diff
@@ -0,0 +1,70 @@
...
+    return formatRemainingTime(date, { locale })
+  }
+
+  const formatOptions: Intl.DateTimeFormatOptions = {
+    day: "2-digit",
+    month: "2-digit",
+    year: "numeric",
+    ...(options?.includeWeekday && { weekday: "long" }),
+    ...(options?.includeTime && { hour: "2-digit", minute: "2-digit" }),
+  }
```
henrikhorluck, line 53: I'm skeptical of this module bearing its weight vs. just telling people to use Intl.DateTimeFormat / Intl.RelativeDateFormat everywhere—it seems it makes more sense to expose this as a set of default format-options, and avoid having to wrap it ourselves?

henrikhorluck (changes requested):

packages/utils/src/formatDate.ts
brage-andreas: I understand where you are coming from. At the same time, I think it simplifies the process to just simply put the date into a function that fixes it for you. I can imagine using Intl as a novice could be needless struggle. I could maybe expose the settings as well, for ""advanced"" date formatting

packages/utils/src/formatDate.ts
```diff
@@ -4,26 +4,40 @@ import { nb, enUS } from "date-fns/locale"
...
+  Time: DEFAULT_INTL_TIME_FORMAT_OPTIONS,
+  Weekday: DEFAULT_INTL_WEEKDAY_FORMAT_OPTIONS,
+  DateTime: { ...DEFAULT_INTL_DATE_FORMAT_OPTIONS, ...DEFAULT_INTL_TIME_FORMAT_OPTIONS },
+  DateWeekday: { ...DEFAULT_INTL_DATE_FORMAT_OPTIONS, ...DEFAULT_INTL_WEEKDAY_FORMAT_OPTIONS },
+  DateTimeWeekday: {
+    ...DEFAULT_INTL_DATE_FORMAT_OPTIONS,
+    ...DEFAULT_INTL_TIME_FORMAT_OPTIONS,
+    ...DEFAULT_INTL_WEEKDAY_FORMAT_OPTIONS,
+  },
+} as const
```
junlarsen, line 29: Enum?

packages/utils/src/formatDate.ts
```diff
@@ -4,26 +4,40 @@ import { nb, enUS } from "date-fns/locale"
...
   "nb-NO": nb,
   "en-US": enUS,
 } as const
 
-type SupportedLocales = keyof typeof dateFNSLocaleFromIntlLocale
+type SupportedLocales = keyof typeof DateFnsLocaleMap
+
+const DEFAULT_INTL_DATE_FORMAT_OPTIONS = { day: "2-digit", month: "2-digit", year: "numeric" } as const
+const DEFAULT_INTL_WEEKDAY_FORMAT_OPTIONS = { weekday: "long" } as const
+const DEFAULT_INTL_TIME_FORMAT_OPTIONS = { hour: "2-digit", minute: "2-digit" } as const
```
junlarsen, line 16: Maybe `satisfies DateTimeFormatOptions`?

apps/web/src/app/events/components/TimeLocationBox/TimeLocationBox.tsx
```diff
@@ -26,13 +26,9 @@ export const TimeLocationBox: FC<Props> = ({
...
-  const [weekday, date, time] = formatDate(datetimeStart, {
-    forceAbsolute: true,
-    includeWeekday: true,
-    includeTime: true,
-  })
-    .replace(",", "")
-    .split(" ")
+  const weekday = Intl.DateTimeFormat("nb-NO", IntlFormats.Weekday).format(datetimeStart)
+  const date = Intl.DateTimeFormat("nb-NO", IntlFormats.Date).format(datetimeStart)
+  const time = Intl.DateTimeFormat("nb-NO", IntlFormats.Time).format(datetimeStart)
```
junlarsen, line 31: prefer new Intl.DateTimeFormat? It's a constructor but it accepts being used as a function

packages/utils/src/formatDate.ts
junlarsen: I think if we want a standard formatting library we should abstract over Intl

packages/utils/src/formatDate.ts
brage-andreas: I'm used to using "as const" over enums. They're lighter and easier to work with in my experience (mostly type-wise). I could move to an enum if you prefer? I see that both are used in the repo

packages/utils/src/formatDate.ts
```diff
@@ -7,13 +7,24 @@ const DEFAULT_LOCALE = "nb-NO"
...
-const DEFAULT_INTL_TIME_FORMAT_OPTIONS = { hour: "2-digit", minute: "2-digit" } as const
+const DEFAULT_INTL_DATE_FORMAT_OPTIONS = {
+  day: "2-digit",
+  month: "2-digit",
+  year: "numeric",
+} as const satisfies Intl.DateTimeFormatOptions
+
+const DEFAULT_INTL_WEEKDAY_FORMAT_OPTIONS = {
+  weekday: "long",
+} as const satisfies Intl.DateTimeFormatOptions
```
junlarsen, line 22: Can remove as const now

packages/utils/src/formatDate.ts
```diff
@@ -26,7 +37,7 @@ export const IntlFormats = {
     ...DEFAULT_INTL_TIME_FORMAT_OPTIONS,
     ...DEFAULT_INTL_WEEKDAY_FORMAT_OPTIONS,
   },
-} as const
+} as const satisfies { [key: string]: Intl.DateTimeFormatOptions }
```
junlarsen, line 40: satisfies Record<string, Intl.DateTimeFormatOptions>

junlarsen (changes requested):

packages/utils/src/formatDate.ts
```diff
@@ -0,0 +1,72 @@
...
+  date: Date,
+  options?: {
+    forceAbsolute?: boolean
+    includeTime?: boolean
+    includeWeekday?: boolean
+    locale?: Intl.LocalesArgument
+    relativeDateThresholdDays?: number
+  }
+): string => {
+  const daysDifference = differenceInDays(new Date(), date)
```
henrikhorluck, line 57: If this is the only place you `date-fns` is used I'd rather replace it with a comparison of unix epochs—this seems to only cause cosmetic "issues", and we don't have to be careful regarding leap-years/timezones/other shenanigans here

packages/utils/src/formatRemainingTime.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Formats the remaining time from the given date.
+ *
+ * @param date - The date to format.
+ * @param options - Optional formatting options.
+ * @param options.locale - The locale to use for formatting. Defaults to `nb-NO`.
+ * @returns The formatted remaining time.
+ */
+export const formatRemainingTime = (date: Date, options?: { locale?: Intl.LocalesArgument }) => {
```
henrikhorluck, line 44: `remainingTime` makes an assumption that `now()` is _before_ `date`, which might not be true

packages/utils/src/formatRemainingTime.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+  if (difference < ONE_YEAR_MS) {
+    return [Math.floor(value / ONE_MONTH_MS), "month"]
+  }
+
+  return [Math.floor(value / ONE_YEAR_MS), "year"]
+}
+
+/**
+ * Formats the remaining time from the given date.
```
henrikhorluck, line 37:
```suggestion
 * Formats the relative time between now and the given date.
```

henrikhorluck (changes requested):
I think we should still encourage direct usage of `Intl.{Date,Relative}TimeFormat`, but these changes LGTM once my last comments are addressed

packages/utils/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "types": "./src/index.ts",
+  "type": "module",
+  "scripts": {
+    "clean": "rm -rf .turbo node_modules",
+    "lint": "biome check . --apply",
+    "lint-check": "biome check .",
+    "type-check": "tsc --noEmit"
+  },
+  "dependencies": {
+    "date-fns": "^2.30.0"
```
henrikhorluck, line 14: Remove this then LGTM

henrikhorluck (approved):

junlarsen (approved):

PR #936, opened 2024-08-13
RFC: Container Service Architecture

junlarsen:
This is a draft for an RFC on our container deployment strategy. Please leave feedback


docs/infrastructure/README.md
```diff
@@ -0,0 +1,167 @@
...
+understand.
+
+The proposed solution has a set of pitfalls outlined in the [Pitfalls & Alternatives](#pitfalls--alternatives) section,
+but we believe that keeping things simple and cheap is rather important for a student organization with limited
+financial resources.
+
+## Service Architecture
+
+The entire application fleet is hosted on a single AWS EC2 (t4g.medium as of writing) instance. The instance is
+available to the public internet through a VPC, but all traffic should be routed from CloudFront to Traefik. Traefik is
```
henrikhorluck, line 33: > all traffic should be routed from CloudFront to Traefik.

Given the intended audience (which includes 1st graders) this should have more ELI5 of _why_ it is smart to route it through CloudFront first

docs/infrastructure/README.md
```diff
@@ -0,0 +1,167 @@
...
+Evergreen instance.
+
+To allow fair share of the instance resources, each container is limited to a subset of the vCPU and memory available on
+the instance. This is configured in the compose configuration file.
+
+The EC2 instance is an Amazon Linux 2023 AMI, and is built with Packer through the terraform monorepo.
+
+In order to allow for instances to be re-deployed or scaled, the Evergreen service is configured behind CloudFront. Each
+time a new Evergreen instance is deployed, the routing table for CloudFront is updated to point to the new instance,
+and the entire CloudFront distribution is invalidated.
```
henrikhorluck, line 49: `docker compose pull && docker compose up --force-recreate` would also work? Seems a bit much to rebuild the entire machine image each deploy if I understand it correctly?

(The document later talks about SSH-ing to do something like this?)

docs/infrastructure/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+**Traefik**
+
+We use Traefik as the routing layer for Evergreen. Traefik is configured with the Docker provider, and uses labels
+available from the Docker socket to automatically discover and load balance containers. Adding new services to Evergreen
```
henrikhorluck, line 54: > load balance containers

Are there _any_ containers we will be running multiple instances of? Why traefik over docker swarm for that case?

docs/infrastructure/README.md
```diff
@@ -0,0 +1,167 @@
...
+- Additional costs for VPC, Route53, other related services
+
+## Pitfalls & Alternatives
+
+Evergreen is a single instance, meaning there is a single point of failure for the entire application fleet. If the
+instance is down, all applications will be down. This should not be a very big problem for a student organization like
+ours. However, it is important to keep in mind that the instance is not intended to be highly scalable, and should
+likely be used for smaller applications that we don't mind small downtimes for.
+
+Whenever we want to scale up (Åre påmelding), we should deploy a new Evergreen instance.
```
henrikhorluck, line 148: I still don't think we need scaling for åre, we just need signup to be async

Literally just introduce a `signup-state: pending`, throw a message on SQS on signup, have a lambda process signups one-by-one

docs/infrastructure/README.md
```diff
@@ -0,0 +1,167 @@
...
+The motivation for Evergreen is to provide a really cheap way to run multiple small applications that can tolerate
+occasional downtime. Additionally, we want to use Docker as the deployment format. Furthermore, Docker Compose is
+relatively simple, and we want to keep deployment complexity low in order to make it easy for new developers to
+understand.
+
+The proposed solution has a set of pitfalls outlined in the [Pitfalls & Alternatives](#pitfalls--alternatives) section,
+but we believe that keeping things simple and cheap is rather important for a student organization with limited
+financial resources.
+
+## Service Architecture
```
henrikhorluck, line 30: Needs a diagram

docs/infrastructure/README.md
junlarsen: Sorry yeah we are only using traefik as a proxy not LB since we wont have multiple containers for one app

docs/infrastructure/README.md
junlarsen: Nothing builds on the ec2 instance, that all happens in gh actions

docs/infrastructure/README.md
junlarsen: Good point

docs/infrastructure/README.md
junlarsen: Updated wording

henrikskog (commented):
Great stuff! Thanks for the detailed writeup. Left a question on the redeployment process.

docs/infrastructure/README.md
henrikskog: Maybe this is easy to implement for ow4? Although all service logic is coupled to django app. But we could just write sql directly in lambda maybe?

docs/infrastructure/README.md
```diff
@@ -0,0 +1,187 @@
...
+
+Each time an application is added to Evergreen, or the docker-compose.yml file is updated, the Evergreen instance must
+be re-deployed. This can be automatically done by Terraform by re-building the Packer AMI, and re-deploying the EC2
+instance with the new AMI.
+
+- Update configuration in Terraform monorepo / docker-compose.yml
+- Re-build Packer AMI
+- Deploy new EC2 instance with new AMI
+- Route CloudFront to the new EC2 instance and invalidate the entire distribution
+- Destroy old EC2 instance
```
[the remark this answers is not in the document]
junlarsen: I can add a detail on it, but each time the compose changes, a new AMI is built with packer and a new EC2 instance deployed

junlarsen:
Partial implementation in progress at https://github.com/dotkom/terraform-monorepo/pull/130

docs/infrastructure/README.md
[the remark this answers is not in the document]
junlarsen: If you are asking if we shouldn't just update the compose on the server is because we want to track changes.

The compose is built into the AMI

PR #939, opened 2024-08-15
Feature/update user-info in settings

madsab:
Closes: #...

## Checklist

- [ ] I have written tests

## Changelog

- ADD: PhoneInput component
- ADD: react-phone-number-input + hookresolver dependencies
- ADD: types/node@* + postcss@^8.3.3 peer dependencies
- ADD: submit form for updating user-info in settings
          - Force format allergies field
          - Show error on submitting of wrong format
          **NOTE:** Because the session does not currently give a full user, i've populated some fields with fake data such as the id of the user. Therefore the only user currently being updated is attached to mads.barnes@hotmail.com

- FIX: Small bugs in Button component and SettingsMenuItem

## How to test
1. Run project
2. Navigate to `settings`
3. Update user

### Default
Screenshot 2024-08-15 at 13 52 37

###  Errors on wrong format
Screenshot 2024-08-15 at 13 52 19

### Success on update user
Screenshot 2024-08-15 at 13 59 08
Screenshot 2024-08-15 at 13 59 33

Check [Auth0](https://manage.auth0.com/dashboard/eu/onlineweb/users/YXV0aDAlN0NjNzVmMmZmYS00YjA3LTQxYzItYjE3My0yNzNhMmI0NDNjNmQ) and [Neon](https://console.neon.tech/app/projects/cool-mode-32262428/branches/br-odd-silence-58530561/tables?database=main) for updated user


apps/web/src/components/views/SettingsView/components/SettingsLanding.tsx
```diff
@@ -1,50 +1,181 @@
...
 
-interface FormInputProps {
-  title: string
-  children?: JSX.Element
-}
+const nameSchema = z
+  .string()
+  .min(2, "Navn må være minst 2 bokstaver")
+  .max(50, "Navn kan maks være 50 bokstaver")
+  .regex(/^[a-zA-ZæøåÆØÅ\s'-]+$/, "Navn kan ikke inneholde tall eller spesialtegn bortsett fra - og '")
```
henrikhorluck, line 17: I don't think we should have a limitation on names here, first of all because regex can cause issues with names from e.g. FEIDE not being valid.

We actually don't need name-validation since we now always set the name from FEIDE, so we have access to users' legal names. Account without membership/usernames should probably only show their emails and we are likely fine with them not being able to change their names

henrikhorluck (changes requested):
I don't think we should support users changing their name given that we now set it from FEIDE

packages/ui/package.json
```diff
@@ -22,13 +22,14 @@
...
     "@radix-ui/react-radio-group": "^1.1.3",
+    "@radix-ui/react-select": "^1.2.2",
     "@radix-ui/react-switch": "^1.0.3",
     "@radix-ui/react-tabs": "^1.0.4",
-    "@radix-ui/react-select": "^1.2.2",
     "clsx": "^2.0.0",
     "cva": "npm:class-variance-authority@^0.7.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
+    "react-phone-number-input": "^3.4.5",
```
henrikhorluck, line 32:
```suggestion
```
Just use `<input type="tel">`? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/tel

apps/web/package.json
```diff
@@ -22,6 +22,7 @@
     "@dotkomonline/types": "workspace:*",
     "@dotkomonline/ui": "workspace:*",
     "@fadi-ui/react-country-flag": "^1.0.7",
+    "@hookform/resolvers": "^3.3.4",
```
henrikhorluck, line 25: What is this used for?

apps/web/src/components/views/SettingsView/components/SettingsLanding.tsx
```diff
@@ -1,50 +1,181 @@
...
+      ...user,
+    },
+  })
+
+  const formatAllergies = (allergies: string) => {
+    setAllergies(
+      allergies
+        .toLowerCase()
+        .replace(/[\s,]+/g, ",")
+        .replace(/(^\w|,\w)/g, (match) => match.toUpperCase())
```
henrikhorluck, line 60: Allergies has to be consistently either a single string-field or a list as presented to users, I don't think we should split it based on commas/spaces, users will just give input that does not turn well into a list

packages/ui/package.json
henrikhorluck: We can also get phone-numbers from FEIDE, though we might need another scope

apps/web/package.json
junlarsen: Zod integration with react-hook-form

apps/web/src/components/views/SettingsView/components/SettingsLanding.tsx
junlarsen: Agree, I don't think we should allow self-name changing since we're also storing the names in two places. More place for error

henrikhorluck: > since we're also storing the names in two places.

What do you mean by this? The single source of truth for us should be Auth0

apps/web/src/components/views/SettingsView/components/SettingsLanding.tsx
madsab: 1. I'm not sure if i understand the problem? If the allergies are used as a string, then most likely all users allergies will be unique and there will be no way of using them for analytics, for example knowing how many attendees in an event are allergic to nuts.
However, it might be a suitable solution to give the user an input-field for ONE allergy, then hitting a `+` to get another input-field for the next allergy.
3. What do you mean by **_users will just give input that does not turn well into a list_**, could you give an example.? In theory, a user can always give bad inputs even if the input is a string or a list

packages/ui/package.json
madsab: 1. If i'm a user who used FEIDE for my account creation, i would not like for the website to get my phone number and show it (this can be a toggle of course). I would rather want to be able to type it in myself, who knows, i might want to use a different phone number than the one registered at FEIDE.

2. By having the country code, not only does it look good in design, but also solves issues with possible members (exchange students) having non-norwegian phone numbers. 

I would like to keep it, but that's my opinion

apps/web/src/components/views/SettingsView/components/SettingsLanding.tsx
madsab: I agree that this can cause issues, but what about users not registering with FEIDE? Can we force name inputs on account registration?

apps/web/src/components/views/SettingsView/components/SettingsLanding.tsx
henrikhorluck: > users allergies will be unique and there will be no way of using them for analytics, for example knowing how many attendees in an event are allergic to nuts.

You would need to perform some kind of text-search w/ synonyms. There is native support for that in e.g. postgres. Also, we don't do such kinds of analytics / statistics, this information is generally only useful on an individual level.

> However, it might be a suitable solution to give the user an input-field for ONE allergy, then hitting a + to get another input-field for the next allergy.

Yes exactly, but from the images in the PR-description it is shown as a text-field, but stored as a list

> could you give an example.?

Look at e.g. the values people have for allergies at old events like https://old.online.ntnu.no/dashboard/events/2268/, there is mix of description and extra contextual information, not just "nuts" / "peanuts" / "pork", and full sentences with weird splits.

IMO: trying to make a regex/split this is any way is a lost cause, just let it be a free text-field, and allow event organizers to sort by non-empty values just read through all of them individually. That is how it currently works, and I think most other features for this is over-engineering or complicating it without giving us any gains. A _potential_ alternative would be that we have a _specific_ list of allergies we _can_ accomodate for, and not allow user-input as text, but IMO that sounds like a lot of work with

- gathering the specific allergies
- makes the user have to actively hunt for their allergies
- maintaining/updating the list

apps/web/src/components/views/SettingsView/components/SettingsLanding.tsx
henrikhorluck: Even social members will have their name set through FEIDE (and if not that is HS badly modifying data / some desperate actions related to ITEX). This is then only relevant for what are essentially guest-users, which I think has only been used for utmatrikulering/immball (?)

packages/ui/package.json
henrikhorluck: > also solves issues with possible members (exchange students) having non-norwegian phone numbers.

Which problems? This dependency seems quite unnecessary, if we wanted to validate that the input data is good we would actually to send a test-sms with a some kind of code akin to 2FA, but that costs money.

> If i'm a user who used FEIDE for my account creation, i would not like for the website to get my phone number and show it (this can be a toggle of course). I would rather want to be able to type it in myself, who knows, i might want to use a different phone number than the one registered at FEIDE.

I agree you might want to register your phone number as something different from FEIDE, but my issue was with adding a whole new dependency which seems unnecessary and just gives us technical debt. I think the feature is fine, just remove the extra JS

henrikhorluck: Regardless of outcome this PR should remove the `@fadi-ui/react-country-flag` dependency

PR #940, opened 2024-08-19
Revert "Disable renovate until it can handle lockfile 9"

henrikhorluck:
Reverts dotkom/monoweb#916


junlarsen:
ah that explains a lot

PR #992, opened 2024-09-05
Don't reset invoicification form when submitting

jotjern:
Bekk representative was complaining that when you click submit on the form seemingly all the form values disappear, making it seem like an empty form was submitted. This PR removes the form reset on submit. Since we redirect on success anyways, it is not necessary to reset the form when successfull, and if the submit fails it is also IMO better to keep the values.

### DIFF
--- a/apps/invoicification/src/app/page.tsx
+++ b/apps/invoicification/src/app/page.tsx
@@ -19,9 +19,8 @@ export default function Page() {
     },
   })
   const dispatch = useSubmitMutation()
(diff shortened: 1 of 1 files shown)


henrikhorluck (changes requested):
Hva slags visuell tilbakemelding får brukeren på at submit-en funker eller ikke funker? Å bare fjerne reset-ingen løser jo ikke problemt de hadde som var at de ikke forstod at det var blitt sendt inn

henrikhorluck (approved):

PR #1003, opened 2024-09-19
Remove tech talks from RIF

brage-andreas:
This PR removes the option to check tech talks from the interest form.


junlarsen (changes requested):
Feel free to just delete the code instead of uncommenting it. We can recover it from git in the future

Also remove the field from the email, react-hook-form useForm() and om the server

junlarsen:
We might mess with the ordering of the checkboxes spreadsheet but I think that's probably ok, maybe just let bedkom know

junlarsen (approved):

PR #1046, opened 2024-10-07
slack email app

henrikskog:
related https://github.com/dotkom/terraform-monorepo/pull/145


apps/email-slack-app/src/core/gmail/InboxMethods.interface.ts
junlarsen, line 1: Is there a reason we have .interface.ts in the file name? Should maybe these types just be a types.ts (or some other relevantly named file)?

apps/email-slack-app/src/lambda.ts
```diff
@@ -0,0 +1,44 @@
+import * as Sentry from "@sentry/aws-serverless"
+import { nodeProfilingIntegration } from "@sentry/profiling-node"
+import type { APIGatewayProxyEventV2, APIGatewayProxyResultV2, Handler } from "aws-lambda"
+import { createExternalClients, createServiceLayer } from "./core"
+import { env } from "./env"
+
+Sentry.init({
+  dsn: "https://a07a847a3bbab26aac0ea417edcbc7f2@o4504113989287936.ingest.us.sentry.io/4508080738467840",
```
junlarsen, line 8: Use an environment variable for this, and disable it in process.env.NODE_ENV !== production

apps/email-slack-app/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../../packages/tsconfig/tsconfig.json",
+  "compilerOptions": {
+    "jsx": "react-jsx"
```
junlarsen, line 4: there's no jsx here right?

junlarsen (commented):
Cool stuff, I suggest we find a more interesting name than email-slack-app (sorry if that causes a lot of terraform updates lol)

Other than that, looks pretty fun

apps/email-slack-app/src/core/gmail/InboxMethods.interface.ts
henrikskog: yeah I agree it's weird, I stripped the code in /gmail from https://github.com/ismail-codinglab/gmail-inbox.

henrikskog: I'll clean up stuff!

apps/email-slack-app/tsconfig.json
henrikskog: Nope thanks

apps/email-slack-app/src/lambda.ts
henrikskog: Yeah good idea, thanks

PR #1053, opened 2024-11-08
Update appkom autobank configuration 

brage-andreas:
See: https://dotkomappkom.slack.com/archives/C071HGV8Q48/p1729252375911129
> Hei dotkom! Kunne dere lagt inn verdier for auth0 client autobank prod på https://github.com/dotkom/monoweb/blob/main/infra/auth0/appkom.tf? Vi trenger det nå i første omgang bare for å kunne bruke ow-apiet til å sjekke en brukers grupper, så verdiene for "allowed_logout_urls" or "callbacks" kan være de samme som i stg :smiley:


junlarsen (approved):

henrikhorluck (changes requested):
They have not given a compelling reason for doing this, why can they not just use dev.onn?

junlarsen:
> They have not given a compelling reason for doing this, why can they not just use dev.onn?

Fair enough but I don't think it's that important either, but then again I think they should probably get their DNS fixed too before anything because we don't really want a .vercel.app in prod

PR #1077, opened 2025-01-23
Register users through RPC call on session creation

jotjern:
TODO:
* Authenticate user before allowing registration


junlarsen (approved):
LGTM, but we need to change so getAuthOptions doesn't depend on core (aka what I said on slack)

packages/auth/src/env.ts
```diff
@@ -0,0 +1,5 @@
+import { createEnvironment, variable } from "@dotkomonline/environment"
+
+export const env = createEnvironment({
+  RPC_HOST: variable,
+})
```
junlarsen, line 5: This should not depend on it, but the app that's calling getAuthOptions should provide ity instead

packages/auth/src/trpc.ts
```diff
@@ -0,0 +1,13 @@
...
+import superjson from "superjson"
+
+export const server = trpc.createTRPCProxyClient<AppRouter>({
+  transformer: superjson,
+  links: [
+    trpc.httpLink({
+      url: `${env.RPC_HOST}/api/trpc`,
+    }),
+  ],
+})
```
junlarsen, line 13: Make it a function that accepts an RPC_HOST

junlarsen (changes requested):

apps/rpc/package.json
```diff
@@ -15,7 +15,7 @@
     "@dotkomonline/db": "workspace:*",
     "@dotkomonline/environment": "workspace:*",
     "@dotkomonline/gateway-trpc": "workspace:*",
-    "@fastify/cors": "^10.0.0",
+    "@fastify/cors": "^9.0.1",
```
junlarsen, line 18: any reason to downgrade cors?

packages/core/src/modules/user/user-repository.ts
```diff
@@ -99,12 +99,19 @@ export class UserRepositoryImpl implements UserRepository {
...
+    const user = await this.getById(auth0Id)
+    if (user === null) {
+      throw new Error("Failed to fetch user after registration")
+    }
+
     await this.db
       .insertInto("owUser")
       .values({ id: auth0Id })
       .onConflict((conflict) => conflict.doNothing())
       .execute()
```
junlarsen, line 112: Shouldn't this be before the other call?

apps/rpc/package.json
jotjern: Not compatible with fastify version

packages/core/src/modules/user/user-repository.ts
jotjern: Why? This makes sure that the user id actually exists before it puts it into the db. getById gets from auth0, not the database

junlarsen: I see

junlarsen (approved):

junlarsen (approved):

PR #1078, opened 2025-01-24
Enable Feide login in auth0

jotjern:
This enables the login with Feide option in auth0 (for ow4 and monoweb)

### DIFF
--- a/infra/auth0/main.tf
+++ b/infra/auth0/main.tf
@@ -132,8 +132,6 @@ resource "auth0_connection" "feide" {
       fetchUserProfile = file("js/fetchUserProfile.js")
     }
   }
-
-  count = terraform.workspace == "prd" ? 0 : 1
 }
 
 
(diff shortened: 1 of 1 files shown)


infra/auth0/main.tf
```diff
@@ -373,6 +371,20 @@ resource "auth0_connection_clients" "username_password_authentication" {
...
+  enabled_clients = [
+    auth0_client.onlineweb_frontend.client_id,
+    auth0_client.onlineweb4.client_id,
+    auth0_client.monoweb_web.client_id,
+    auth0_client.monoweb_dashboard.client_id,
+    auth0_client.vengeful_vineyard_frontend.client_id,
+    auth0_client.appkom_opptak.client_id,
+    auth0_client.appkom_events_app.client_id,
+    auth0_client.appkom_autobank.client_id,
+  ]
```
junlarsen, line 385: What does this imply? That all of the following apps can use Feide as a provider?

I think to begin with it's fair to keep first-party apps only?

jotjern: These are all the apps which can use feide to log in (to the best of my understanding), meaning that if we want to make feide-login the default we have to enable it everywhere people use onlineweb authentication.

junlarsen (approved):

PR #1106, opened 2025-01-25
Make migrator use doppler environment monoweb-migrator by default

jotjern:
This makes it possible to run pnpm migrate for the dev environment without exporting DATABASE_URL or other manual setup


junlarsen (changes requested):
How do you override? Pls add docs

junlarsen:
sry for dårlig review jeg er full

PR #1142, opened 2025-02-01
Readme file

sondrealf:
Updated readme to main readme file and contributed.

Think how to run locally should be in the main readme file to keep it simple for beginners.


junlarsen (approved):
Dont need two terminals if you use docker compose up -d

PR #1143, opened 2025-02-01
Do not directly access client variables without createEnvironment

junlarsen:
Reverts two prior commits and passes the arguments as parameters to createEnvironment as intended.

This keeps the type safety that we intended for the environment package

cc @jotjern


jotjern (approved):
lgtm

PR #1145, opened 2025-02-02
Upgrade to Next 15

jotjern:
Upgrade to next 15. Done by:
* Removing swcMinify: true from all next configs because it is now the default
* Using await on the cookies() method from "next/headers"
* Removing next-superjson-plugin because it is experimental and crashes.
* Replace all usage of url params in pages with async variant
* Fix hydration errors for color themes in web and dashboard
* Fix invalid key props in a couple of places

Note:
The next superjson plugin is almost of no use after the upgrade from pages to app router.
https://www.npmjs.com/package/next-superjson-plugin
The only use case is when passing a date to a client component, but even then you need to use the "data-superjson" tag, which is not worth having in my opinion.


apps/dashboard/src/app/(dashboard)/interest-group/[id]/layout.tsx
```diff
@@ -5,11 +5,12 @@ import { type PropsWithChildren, useMemo } from "react"
 import { trpc } from "../../../../trpc"
 import { InterestGroupDetailsContext } from "./provider"
 
-export default function InterestGroupDetailsLayout({
+export default async function InterestGroupDetailsLayout({
```
Comment (junlarsen) [inline apps/dashboard/src/app/(dashboard)/interest-group/[id]/layout.tsx:8]: Isn't this illegal? Async use client?

apps/web/next.config.mjs
```diff
@@ -14,9 +13,6 @@ const config = {
       },
     ],
   },
-  experimental: {
-    swcPlugins: [["next-superjson-plugin", {}]],
-  },
```
junlarsen, line 19: So passing date/map etc works out of the box?

jotjern: No, it does not work. But to use it, even with next-superjson-plugin you have to use the "data-superjson" prop, which we don't use anywhere, and the only case where it is useful is passing data from server to client props which is imo marginal. The plugin does not support next 15.

junlarsen (changes requested):
Still a bunch of async use clients here that I think needs to be rewritten to use()

junlarsen (approved):
Nice! Do you want to create a PR for upgrading to React 19 as well?

jotjern:
> Nice! Do you want to create a PR for upgrading to React 19 as well?

tRPC does not yet support react 19, at least not the stable version (10.x). Tried looking for some indications on when tRPC 11.x is set to be released, the beta has been out for over a year now, but didn't find much

PR #1148, opened 2025-02-03
Add tests for corrupt migrations

jotjern:
Adds tests that ensure migrations are reversible by running each migration up and down and checking that the database tables and types are the same after. 

Adds a CI step which validates migrations


packages/db/package.json
```diff
@@ -9,17 +9,21 @@
     "clean": "rm -rf .turbo node_modules",
     "lint": "biome check . --write",
     "lint-check": "biome check .",
-    "type-check": "tsc --noEmit"
+    "type-check": "tsc --noEmit",
+    "test": "vitest run"
   },
   "dependencies": {
     "@dotkomonline/logger": "workspace:*",
+    "@testcontainers/postgresql": "^10.13.2",
```
junlarsen, line 17: These should be devdependencies

Also maybe consider dropping chalk, testfailure text diff is already red right?

packages/db/src/migrations.test.ts
```diff
@@ -0,0 +1,102 @@
...
+  const result = await kysely.executeQuery<{ typname: string }>(
+    sql`SELECT typname FROM pg_type t WHERE typcategory = 'E'
+  `.compile(kysely)
+  )
+
+  return result.rows.map((row) => row.typname)
+}
+
+async function resetTestDatabase(kysely: Kysely<Database>) {
+  // This is a dangerous operation, so we want to make sure we are in the right database
```
junlarsen, line 17: Good call

junlarsen (changes requested):

junlarsen (approved):
lgtm

PR #1152, opened 2025-02-04
Server side authentication

jotjern:
This implements server side authentication by doing some type hackery to trick next-auth into thinking we have a request object. This can be replaced by something more sensible when next-auth 5 is released, but for now it works. Read the comment in the PR, you can verify that this works by reading the source code linked.


apps/web/src/utils/trpc/server.ts
```diff
@@ -1,14 +1,48 @@
 import { env } from "@/env"
 import type { AppRouter } from "@dotkomonline/gateway-trpc"
 import * as trpc from "@trpc/client"
+import { getToken } from "next-auth/jwt"
+import type { RequestCookies } from "next/dist/compiled/@edge-runtime/cookies"
+import type { ReadonlyHeaders } from "next/dist/server/web/spec-extension/adapters/headers"
+import type { ReadonlyRequestCookies } from "next/dist/server/web/spec-extension/adapters/request-cookies"
```
junlarsen, line 7: these paths are a little sus but seems ok for now. hopefully we can get rid of all of this with next auth 5

junlarsen (approved):
lgtm

junlarsen:
Fails to build here

image

jotjern:
> Fails to build here

Static pages which don't have cookies and headers available, added try catch. Bit ugly 🤷‍♂️

Issue #235, opened 2022-11-15
Replace prisma with kysely, or another sql builder


kharann:
### ❓ Is your feature request related to a problem? Please describe.

Prisma is a great library with good developer experience that speeds up the development process a lot. However, in our case, it might not be the best fit. It has some issues that has been slow to fix.

1. [No schema splitting](https://github.com/prisma/prisma/issues/2377). Big files will become problematic in the future and should have first-class support. There are third-party tools, but any of them might break if the prisma-team introduces breaking changes. 
2. [No programmatic access to CLI](https://github.com/prisma/prisma/issues/4703). Prisma uses some kind of binary that makes things difficult to migrate programmatically. Current solutions include spawning child processes. Programmatic access is nice for migrating inside the cloud and dynamically migrating test contains etc. 
3. [Lambda cold start](https://github.com/prisma/prisma/issues/10724). The library is not very small and introduces some cold start in a serverless setting

### 📝 Describe the solution you'd like to see

The rest of the ORMs in this space seems to be kinda lackluster. Perhaps we should look at query builders

- Knex
- Slonik
- Kysely

Out of all of these, kysely is the youngest but seems to have pretty cool type inference and DX. Since it's just SQL tbh, and uses the node `pg` under the hood, it seems to be safe to bet on it for now.

### 💡 Describe alternate solutions

Back to raw SQL is also an option.

PR #1168, opened 2025-02-12
Replace Kysely with Prisma for database interaction

jotjern:
This pull request replaces Kysely with Prisma. This is done for a couple reasons:
* Automatic migrations: It currently takes considerable effort to do minor changes to the database structure
* Improved typescript performance: kysely has poor typescript performance compared to prisma
* Generated types: currently we define all of our types doubly, both in the types module and in the migrations. This PR still allows deviation from the db schema by extending the generated schema, but they are based on the db schemas.

Ref some points from earlier: https://github.com/dotkom/monoweb/issues/235
1. **No schema splitting**
This is now possible, but I don't see why this is a big issue anyways, having a large file is fine.
2. **No programmatic access to CLI**
We can wrap it, it works just fine
3. **Lambda cold start**
The Prisma team have rewritten the rust part of the codebase, so the large binaries shouldn't be a problem anymore.

This change drastically reduces the number of steps necessary to make minor changes by simply using `schema.prisma` as source of truth, and then generating migrations and domain object attributes based on those. 

Prisma is a very well maintained project, and is in my opinion well suited for monoweb.

The pull request is quite large and ugly because of all of the repository changes, those are most of the changes. The only non-trivial service change is job-listing-service which i refactored because the current solution was very much broken.


packages/db/eu-north-1-bundle.pem
junlarsen, line 1: Delete this file?

junlarsen (changes requested):

apps/dashboard/src/app/(dashboard)/article/[id]/layout.tsx
```diff
@@ -13,7 +13,7 @@ export default function ArticleDetailsLayout({
   const { data, isLoading } = trpc.article.get.useQuery(id)
   const value = useMemo(
     () =>
-      data === undefined || isLoading
+      !data || isLoading
```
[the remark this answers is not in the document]
Comment (jotjern) [reply apps/dashboard/src/app/(dashboard)/article/[id]/layout.tsx:16]: Yes. Undefined when the query is loading and null if the article was not found

packages/core/src/modules/article/article-service.ts
```diff
@@ -68,7 +68,7 @@ export class ArticleServiceImpl implements ArticleService {
       throw new ArticleNotFoundError(id)
     }
     let name = await this.articleTagRepository.getByName(tag)
-    if (name === undefined) {
+    if (name === null) {
```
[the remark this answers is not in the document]
jotjern: This refactor uses null for every case.

packages/db/eu-north-1-bundle.pem
jotjern: This file is needed, prisma needs to have a file path in the database url instead of passing it via environment variable.

doppler.yaml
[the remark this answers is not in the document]
jotjern: Formatting that ran, but there is an actual change, removing the migrator

henrikskog:
😱

apps/web/src/components/views/SettingsView/components/SettingsLanding.tsx
```diff
@@ -57,13 +57,13 @@ const Landing: NextPage<{ user: User }> = ({ user }) => {
           <TextInput
             width="flex-1 mb-2 mx-1"
             placeholder="Fornavn"
-            defaultValue={user.firstName}
+            defaultValue={user.firstName ?? ""}
```
junlarsen, line 60: Questionable, should it not be `?? undefined`?

packages/core/package.json
```diff
@@ -31,11 +31,10 @@
   "devDependencies": {
     "@biomejs/biome": "^1.9.3",
     "@dotkomonline/tsconfig": "workspace:*",
+    "@testcontainers/postgresql": "^10.18.0",
     "@dotkomonline/types": "workspace:*",
-    "@testcontainers/postgresql": "^10.13.2",
     "@types/node": "^22.0.0",
     "@vitest/ui": "^1.3.1",
-    "testcontainers": "^10.13.2",
```
junlarsen, line 38: Why was this removed?

packages/core/src/modules/attendance/attendance-repository.ts
```diff
@@ -22,53 +20,71 @@ export interface AttendanceRepository {
...
+        },
+      },
+    })
+
+    if (attendance === null) return null
+
+    return this.parseExtras(attendance)
+  }
+
+  // Prisma requires distinction between database null and json null, so here we choose database null
```
junlarsen, line 77: Use JSDoc comments if above function definition, or move into body for regular comment

packages/core/src/modules/attendance/attendance-repository.ts
```diff
@@ -22,53 +20,71 @@ export interface AttendanceRepository {
...
+  private correctNullTypes<T extends { extras?: unknown }>(data: T) {
+    return { ...data, extras: data.extras === null ? Prisma.DbNull : data.extras }
+  }
+
+  // Takes an object with unparsed JSON value yearCriteria and returns it with yearCriteria parsed
```
junlarsen, line 82: Same here

packages/core/src/modules/job-listing/job-listing-location-link-repository.ts
junlarsen, line 1: Assume you merged this with job-listing-repository?

packages/core/src/modules/job-listing/job-listing-location-repository.ts
junlarsen, line 1: Assume you merged this with job-listing-repository?

packages/core/src/query.ts
```diff
@@ -1,98 +1,22 @@
...
-  }
-
-  const data = records.slice(0, pageable.take).map(mapper)
   return {
-    next: cursor,
-    data,
-    count: Math.min(pageable.take, data.length),
+    cursor: { id: page.cursor },
+    orderBy: { id: "desc" as const },
+    skip: page.cursor ? 1 : 0,
```
junlarsen, line 17: This is fairly unsound (not that it chagned much in this PR) because UUIDs are not monotonically sortable.

We will have to change cursor to be a tuple (id, created_at) in a later PR

packages/core/vitest-integration.setup.ts
```diff
@@ -1,84 +1,69 @@
...
+beforeEach(async () => {
+  console.log("before:", await core.jobListingService.getLocations())
+
+  await dbClient.$queryRaw`BEGIN TRANSACTION;`
+})
+
+afterEach(async () => {
+  await dbClient.$queryRaw`ROLLBACK;`
+
+  console.log("after:", await core.jobListingService.getLocations())
```
junlarsen, line 68: Could also just create dummy databases for each test, but I guess this is ok

junlarsen: Remove the console.logs however

packages/db/.env
junlarsen, line 1: Does this file have to be here?

packages/db/.gitignore
junlarsen, line 1: It contradicts the gitignore rule here

packages/db/prisma/schema.prisma
```diff
@@ -0,0 +1,403 @@
+generator client {
+  provider = "prisma-client-js"
+}
+
+generator zod {
+  provider = "zod-prisma-types"
+  output   = "../src/generatedSchemas"
```
junlarsen, line 7: Keep file path naming consistent
```suggestion
  output   = "../src/generated"
```
Or generated-schemas

packages/db/prisma/schema.prisma
```diff
@@ -0,0 +1,403 @@
...
+  PENDING
+  APPROVED
+  REJECTED
+}
+
+enum CompanyType {
+  Consulting
+  Research
+  Development
+  Other
```
junlarsen, line 58: Shouldn't this be all UPPERCASE like the other enums too?

packages/db/prisma/schema.prisma
junlarsen, line 1: Can you `@@map()` each table/relation so they follow standard postgres naming? snake_case everything?

Last time I used prisma you would receive wildly inconsistent table/link table names

junlarsen, line 1: There is no good reason to use db.varchar over just TEXT in postgres. It just creates hassle if you need to change the constraints in the future. These are better written in the application code than the DB schema

Postgres does not differentiate the storage methods of TEXT vs VARCHAR, they are both stored using the same internal representation so there is no reasonable performance difference either

packages/db/prisma/schema.prisma
```diff
@@ -0,0 +1,403 @@
...
+  events      EventCompany[]
+  JobListing  JobListing[]
+}
+
+model Committee {
+  id          String           @id @default(uuid())
+  createdAt   DateTime         @default(now())
+  name        String           @db.VarChar(100)
+  description String           @default("")
+  email       String           @default("kontakt@online.ntnu.no")
```
junlarsen, line 104: Strongly disagree with this default

packages/db/prisma/schema.prisma
```diff
@@ -0,0 +1,403 @@
...
+  image       String?
+  events      EventCompany[]
+  JobListing  JobListing[]
+}
+
+model Committee {
+  id          String           @id @default(uuid())
+  createdAt   DateTime         @default(now())
+  name        String           @db.VarChar(100)
+  description String           @default("")
```
junlarsen, line 103: Use NULL as default instead of empty strings

packages/db/prisma/schema.prisma
```diff
@@ -0,0 +1,403 @@
+generator client {
+  provider = "prisma-client-js"
+}
+
+generator zod {
+  provider = "zod-prisma-types"
+  output   = "../src/generatedSchemas"
+
+  useMultipleFiles = true
```
junlarsen, line 9: I would probably prefer to have all the codegen in a single file. It's less files to search through in text editors

packages/db/prisma/schema.prisma
```diff
@@ -0,0 +1,403 @@
...
+model InterestGroup {
+  id              String   @id @default(uuid())
+  createdAt       DateTime @default(now())
+  updatedAt       DateTime @updatedAt
+  name            String
+  description     String
+  link            String?
+  isActive        Boolean  @default(true)
+  longDescription String   @default("")
+  joinInfo        String   @default("")
```
junlarsen, line 391: Make them nullable. Also why is joinInfo a different field? I think it should all just be one MDX blob

packages/db/src/fixtures.ts
```diff
@@ -0,0 +1,54 @@
...
+import { getCompanyFixtures } from "./fixtures/company"
+import { getEventFixtures } from "./fixtures/event"
+import { getInterestGroupFixtures } from "./fixtures/interest-group"
+import { getJobListingFixtures, getJobListingLocationFixtures } from "./fixtures/job-listing"
+import { getMarkFixtures } from "./fixtures/mark"
+import { getOfflineFixtures } from "./fixtures/offline"
+import { getProductFixtures } from "./fixtures/product"
+import { getProductPaymentProviderFixtures } from "./fixtures/product-payment-provider"
+
+if (process.env.DATABASE_URL === undefined) {
```
junlarsen, line 15: QOL: Add a if DATABASE_URL is pointing to AWS, then abort

packages/db/src/generatedSchemas/inputTypeSchemas/WaitlistAttendeeScalarFieldEnumSchema.ts
junlarsen, line 1: The generated schemas should not be commited to version control.

packages/db/src/index.ts
```diff
@@ -1,42 +1,10 @@
...
-import pg, { type PoolConfig } from "pg"
-import type { DB } from "./db.generated"
-export type { DB as Database } from "./db.generated"
-import { getLogger } from "@dotkomonline/logger"
+import { type Prisma, PrismaClient } from "@prisma/client"
+import type { DefaultArgs } from "@prisma/client/runtime/library"
 
-export { createMigrator } from "./migrator"
-const logger = getLogger("@dotkomonline/db")
+export * as dbSchemas from "./generatedSchemas"
```
junlarsen, line 4: I would prefer if it's just named schemas. I think it's quite unmistakenable what schemas we're talking about here

```suggestion
export * as schemas from "./generatedSchemas"
```

junlarsen (changes requested):

packages/db/prisma/schema.prisma
jotjern: joinInfo is used by appkom opptak

packages/db/prisma/schema.prisma
jotjern: IMO not worth enabling preview features for such a small improvement. Also, since it's at the top of the file I think it's easy to find. But agree we should use it when they release it

packages/db/src/fixtures.ts
jotjern: We most likely want to run this on staging which is on AWS. I'll add a check for prod

packages/core/src/modules/job-listing/job-listing-location-link-repository.ts
jotjern: Yes, the repository worked by a bug, so it broke when I reworked it for prisma.

packages/core/package.json
jotjern: I don't think it's needed, it runs just fine without it. I've reworked the code here anyways

packages/core/src/query.ts
```diff
@@ -1,98 +1,22 @@
...
-    builder = builder.where(sql`id`, "<", sql`${pageable.cursor.id}`)
-  }
-  // Always perform ordering on id.
-  builder = builder.orderBy(sql`id`, "desc")
-
-  // Perform the query and build the output
-  const records = await builder.execute()
-  const hasNextCursor = records.length !== 0 && records.length === pageWithNextLength
-  const last = records.at(-1)
+export function pageQuery(page: Pageable): { orderBy: { id: "desc" }; cursor?: { id: string }; skip?: number } {
```
junlarsen, line 11: I don't like this return type, but I'm OK with it for now.

packages/db/src/fixtures.ts
junlarsen: I think staging should just hold an older version of data from prod.

junlarsen (changes requested):
Mostly happy with the changes now, still have some comments on the schema file itself.

packages/db/src/fixtures.ts
jotjern: In the future yes, right now we need test data for staging and prod does not have it.

packages/db/src/testDatabases.ts
junlarsen, line 1: rename file to `test-databases` for consistency

packages/db/prisma/schema.prisma
junlarsen: Still missing this

junlarsen (changes requested):

packages/db/prisma/schema.prisma
jotjern: Whoops, missed this

packages/db/.env
jotjern: This is the best way I found to make the db migrate the local database by default

packages/db/.gitignore
jotjern: I think it is good that changes to it are not tracked, as someone could put secrets in it. Not an ideal solution, but IMO not that bad either. Maybe change later?

junlarsen (approved):
I think this one is ready to go now, well done :D

PR #1205, opened 2025-02-23
Replace Committee with Group

Andrefkl:
* Rename Committee to Group and add GroupType for Committees, NodeCommittees and OtherGroups
* Add Groups to dashboard


jotjern (approved):
leste ikke men fikk popcorn

PR #1206, opened 2025-02-24
Move typescript config into @dotkomonline/config

junlarsen:
Trying to slim down the number of packages in the monorepo


brage-andreas (approved):
Nice job! 👍

PR #232, opened 2022-11-10
DRAFT: Kysely

kharann:
Closes: #...

## Checklist

- [ ] I have written tests
- [ ] I have provided documentation

## Changelog

- A bullet list of what you've done to make the life of the reviewer easier.

## How to test

A consise description of how to test the newly implemented functionality.

PR #1208, opened 2025-02-25
Ensure docker images can build in pipeline

junlarsen:
The docker build is now tested in CI

This introduces some internal changes, explained in review comments


apps/rpc/runtime.mjs
```diff
@@ -0,0 +1,21 @@
+import path from "node:path"
+import { URL } from "node:url"
+
+export async function resolve(specifier, context, nextResolve) {
```
junlarsen, line 4: Rewrites `../index` to `../index.ts` for node imports. This means things work well with both next's build system and node --strip-experimental-types

packages/db/package.json
```diff
@@ -2,9 +2,21 @@
   "name": "@dotkomonline/db",
   "version": "1.0.0",
   "type": "module",
-  "main": "./src/index.ts",
-  "types": "./src/index.ts",
   "private": true,
+  "exports": {
```
junlarsen, line 6: Needed, so that `types` only imports schemas, and none of the actual Prisma client stuff. The prisma client breaks on Vercel (and shouldn't be used there for any reason either).

Same goes for the test harness

packages/db/prisma/schema.prisma
```diff
@@ -6,7 +6,7 @@ generator zod {
   provider = "zod-prisma-types"
   output   = "../src/schemas"
 
-  useMultipleFiles = true
+  useMultipleFiles = false
```
junlarsen, line 9: Node does not support directory import, and this codegenerator generates directory import when using multiple files.

packages/db/src/index.ts
```diff
@@ -1,10 +1,15 @@
...
+const require = createRequire(import.meta.url)
+const { Prisma: _Prisma, PrismaClient: _PrismaClient } = require("@prisma/client")
+export type * from "@prisma/client"
 
+export const PrismaRuntime = _Prisma
+export const PrismaClientRuntime = _PrismaClient
 export type DBClient = PrismaClient<Prisma.PrismaClientOptions, never, DefaultArgs>
 export const createPrisma = (databaseUrl: string): DBClient =>
-  new PrismaClient({
+  new _PrismaClient({
```
junlarsen, line 13: Prisma does not generate ESM module code, so we have to createRequire() and require it here. For this reason, the `Prisma` export cannot be both the client and namespace (like the generated client does).

Therefore we have to distinguish between Prisma namespace (re-exported from the client), and the runtime values.

junlarsen: This means the Dockerfile build was broken in #232, but we never noticed because CI never built them. We'd have to do this at some point anyways

brage-andreas (approved):

PR #1219, opened 2025-03-05
Add dockerfile build for web

junlarsen:
Builds web with rpc host/public hostname build arguments


apps/web/next.config.mjs
```diff
@@ -21,6 +22,7 @@ const config = {
     "@dotkomonline/types",
     "@dotkomonline/ui",
   ],
+  ...(isStandaloneBuild ? { output: "standalone" } : {}),
```
junlarsen, line 25: Standalone build switch so that we can still build vercel in preview

PR #1225, opened 2025-03-07
Add backend code for kvittering

henrikskog:
CD setup needs testing


.github/workflows/deploy-prod.yml
```diff
@@ -0,0 +1,44 @@
+name: Production deployment
+on:
+  push:
+    branches:
+      - main
```
junlarsen, line 5: This should probably be changed to on release later, but this is OK for now

.github/workflows/deploy-prod.yml
```diff
@@ -0,0 +1,44 @@
...
+      - uses: aws-actions/amazon-ecr-login@v2
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+      - uses: docker/build-push-action@v6
+        with:
+          platforms: linux/amd64
+          context: .
+          file: apps/kvittering-backend/Dockerfile
+          tags: 891459268445.dkr.ecr.eu-north-1.amazonaws.com/monoweb/kvittering-backend:latest
```
junlarsen, line 28: These should be per-environment, so monoweb/prod/kvittering-backend, you will probably need infra changes here too

apps/kvittering-backend/README.md
junlarsen, line 1: Remove empty file

apps/kvittering-backend/src/core/email_service.py
```diff
@@ -0,0 +1,109 @@
...
+    def _create_default_ses_client():
+        """Create and return a default Amazon SES client."""
+        try:
+            return boto3.client('ses', region_name="eu-north-1")
+        except Exception as e:
+            logger.error(f"Failed to create SES client: {str(e)}")
+            raise Exception(f"Could not initialize email service: {str(e)}")
+    
+    @staticmethod
+    def get_formatted_text(form: FormData) -> str:
```
junlarsen, line 34: Would be cool we if we could use our emailing template service here in the future

(That one doesn't have code for attachments at the moment so I don't think it's possible to use right now)

junlarsen (changes requested):
Haven't looked too much into the python code, but it seems alright. Needs some changes on the infra side, and it would be nice to document how this thing runs either inside /docs or in the readme.

.github/workflows/deploy-prod.yml
henrikskog: True, nice catch

apps/kvittering-backend/src/core/email_service.py
henrikskog: Yeah I considered using brevduen but no attachments as you say, and it was quicker to just fix it myself here

henrikskog:
Cd kjører:

Commit: https://github.com/dotkom/monoweb/pull/1225/commits/829aaceea7fd2d1078bf798b5091732c45b5554b
https://github.com/dotkom/monoweb/actions/runs/13723438949/job/38384133869?pr=1225
https://github.com/dotkom/terraform-monorepo/actions/runs/13723454501/job/38384187505

PR #1236, opened 2025-03-07
kvittering frontend

henrikskog:
infra: https://github.com/dotkom/terraform-monorepo/pull/174/files


junlarsen:
rebase on main so backend changes dont show in the diff

pnpm-workspace.yaml
```diff
@@ -2,3 +2,6 @@ packages:
   - apps/*
   - packages/*
   - tools/*
+
+  - '!apps/monoweb-kvittering-backend' # python app
+  - '!apps/monoweb-kvittering-frontend' # does not use any monorepo packages as of now, so we exclude it
```
junlarsen, line 7: Should be added here, its wrong not to because otherwise it gets its own lock

junlarsen (changes requested):
Do we need all these scripts? Doesnt this deploy on its own?

Delete ngunx/docker compose

pnpm-workspace.yaml
junlarsen: Because then it's effectively not a part of the monorepo whoch kind of defeats the purpose of a monorepo

henrikskog:
Fucked up while rebasing on main rip. And something weird happening with this kvittering-frontend branch that I dont understand locally, so the easiest was to create a new branch and new PR  https://github.com/dotkom/monoweb/pull/1245

pnpm-workspace.yaml
henrikskog: and yep ill remove all the scripts, not needed

henrikskog: Hmm sure I was thinking that as its using any of the packages it does not really matter

PR #1296, opened 2025-03-19
Bootstrap splash 2025 project

henrikskog:


junlarsen:
This should not be its own app

PR #1311, opened 2025-03-20
Change Text and Title components to be polymorphic

junlarsen:
Also removes xs/xl/2xl Text sizes since they were completely unused and
because we don't need that many body scales.


junlarsen:
### Merge activity

* **Mar 20, 7:06 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/dotkom/monoweb/1311).

PR #1331, opened 2025-03-30
Fix bug where brand color default was brand-3

henrygraesberg:
The default for brand color in the tailwind preset was brand-3, whereas the actual brand color for Online (#0D5474) was brand-9


junlarsen (changes requested):
All of them should actually default to 9 according to https://www.radix-ui.com/colors/docs/palette-composition/understanding-the-scale

junlarsen (approved):

PR #1370, opened 2025-04-06
Create OAuth2 authentication library

junlarsen:
# Monoweb OAuth2 Library

This PR introduces an oauth2 library (`packages/oauth2`) that doesn't make any stupid assumptions. It's built specifically for Auth0 so we don't suffer the generalizations that next-auth makes. The library consists of the following components:

1. OAuth2/Auth0 compliant authentication library
2. JWT Verification service (previously lived in packages/gateway-trpc, now moved here)
3. React context library
4. Next.js `getServerSession` and route handlers for authorize/logout/callback/session endpoints
5. Session utilities

## Changes

- Introduced new OAuth2 library
- Rewrote web/dashboard to use this library
- Changed web/dashboard to call the RPC service directly (accessToken now available on client)
- Removed next-auth and packages/auth
- Removed packages/proxy-nextjs
- Sister PR #1371 adds logout urls for web/dashboard
- Updated all dependent code to use the new library and its APIs

## Motivation

This library aims to fix the problems we've ran into using next-auth.

Next-auth is made to be provider-agnostic and makes a lot of assumptions in its types about what the tokens have or do not have. It's also tricky to understand its callback API because it behaves wildly different whether you use JWT sessions or not.

It also has zero built-in support for refresh token rotation which means you have to hack the framework anyways.

A lot of its API surface is built around having Account/User distinction, but we don't care for that at all. It also makes a distinction between User and Session, and the difference between session/jwt age. Our library makes these the same.

## Authentication & Session management

We currently do a little dubious session management by storing the session in a signed JWT sent to the browser. This does have some problems (e.g you cannot revoke sessions) but it's the easiest (no database required) and it's what our previous next-auth config also did. It should be very easy to switch this out in the future.

A client can initiate auth by redirecting the browser to `/api/auth/authorize` which will redirect to Auth0, and stuff flows like regular OAuth2. Logout endpoints also work given that the Auth0 client has logout urls configured.

## Roadmap

The OAuth2 library has code for refresh token rotation, but none of the frontends use this at the moment.

The callback does currently not validate the `nonce` claim on the ID token.


junlarsen:
* **#1370** <a href="https://app.graphite.dev/github/pr/dotkom/monoweb/1370?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/dotkom/monoweb/1370?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jotjern (approved):
lgtm

PR #1389, opened 2025-04-10
Get environment variables for build dynamically in CI

jotjern:


junlarsen:
#1396 introduces https://github.com/dotkom/monoweb/blob/users/junlarsen/try_composite_action_for_web_build/.github/actions/internal/doppler-secrets/action.yml which both exports them to github env and as output for build args.

PR #1425, opened 2025-04-23
Implement attendee list

ErlendSae:
image
Looking like this now

### DIFF
--- a/apps/web/src/app/events/components/AttendanceCard/AttendeesList.tsx
+++ b/apps/web/src/app/events/components/AttendanceCard/AttendeesList.tsx
@@ -1,5 +1,5 @@
 import { useTRPC } from "@/utils/trpc/client"
-import type { AttendancePool, Attendee } from "@dotkomonline/types"
+import type { Attendee } from "@dotkomonline/types"
 import { Avatar, AvatarFallback, Icon, Text } from "@dotkomonline/ui"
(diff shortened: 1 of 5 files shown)


apps/web/src/app/events/components/AttendanceCard/AttendeesList.tsx
```diff
@@ -0,0 +1,77 @@
...
+            <div className="text-sm text-slate-11">
+              {displayPools.find((p: AttendancePool) => p.id === vipAttendee.attendancePoolId)?.title}
+            </div>
+          </div>
+        </div>
+      )}
+      */}
+
+      <div className="flex flex-col gap-2">
+        <h3 className="font-medium">Påmeldte</h3>
```
junlarsen, line 51: Use components from `@dotkomonline/ui`

junlarsen: Also probably replace one (or multiple of the divs) below with the Text component

ErlendSae:
image

junlarsen:
Looks nice. I think either one is fine. Maybe the question to ask is how does the user benefit or not when seeing the groups vs not seeing the groups

brage-andreas:
Not perfect but should work for now
image

PR #1427, opened 2025-04-23
Mono 365 add a delete event button in dashboard

ErlendSae:
image
image


apps/dashboard/src/app/form.tsx
```diff
@@ -460,6 +461,7 @@ interface FormBuilderOptions<T extends z.ZodRawShape> {
   }>
   defaultValues?: DefaultValues<z.infer<z.ZodObject<T>>>
   label: string
+  extraButtons?: ReactNode
```
junlarsen, line 464: Just a nit, but maybe call these `actions`?

apps/dashboard/src/app/form.tsx
```diff
@@ -506,7 +509,10 @@ export function useFormBuilder<T extends z.ZodRawShape>({
         <Flex direction="column" gap="md">
           {components}
           <div>
-            <Button type="submit">{label}</Button>
+            <Group>
+              <Button type="submit">{label}</Button>
```
junlarsen, line 513: If I'm not mistaken, this Button needs a `key` prop since it's next to the extra buttons

packages/core/src/modules/event/event-repository.ts
```diff
@@ -122,4 +123,8 @@ export class EventRepositoryImpl implements EventRepository {
   async removeEventFromInterestGroup(eventId: EventId, interestGroupId: InterestGroupId): Promise<void> {
     await this.db.eventInterestGroup.delete({ where: { eventId_interestGroupId: { interestGroupId, eventId } } })
   }
+
+  async delete(id: EventId): Promise<void> {
+    await this.db.event.delete({ where: { id } })
+  }
```
junlarsen, line 129: We can get to this later because it requires more code changes, but we should ideally never truly delete data.

There should probably be a `deleted: DateTime?` field on the event object, where `null` signals its active, and a time value tells the time it was "deleted"

junlarsen (commented):
Nice, this is definitely going to be needed when arrkom gets in here :wheelchair: 

I added a small nit on some naming, and a note on how we should "delete" events.

apps/dashboard/src/app/form.tsx
ErlendSae: The group doesnt render a list and is a layout container, so i dont think its necessary.

jotjern (changes requested):
I have implemented soft deletes now, so you just have to update it with status: "DELETED"

PR #1442, opened 2025-04-24
Implement soft deletion of events

jotjern:
ai generert description lol men den var riktig

### TL;DR

Updated event status options to use DRAFT, PUBLIC, and DELETED instead of the previous status options.

### What changed?

- Changed event status enum values from TBA/PUBLIC/NO_LIMIT/ATTENDANCE to DRAFT/PUBLIC/DELETED
- Updated form dropdowns in both edit-form.tsx and write-form.tsx to reflect the new status options
- Changed default event status from TBA to DRAFT
- Added filters in event repository to exclude DELETED events from queries
- Created a database migration to update the event_status enum
- Updated fixture data to use the new status values


jotjern:
> We should have a record of when something was deleted, so there should be a timestamp on the type

We plan to implement an audit log, so I don't think it will be necessary to add this

junlarsen (approved):
Alright, careful when applying the migration to prod :thinking: it's removing the entire column for some reason

jotjern:
I manually wrote the migration so it would just default everything to be public, the old statuses were nonsensical so I didn't think there was any value in them

jotjern:
### Merge activity

* **Apr 25, 10:33 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/dotkom/monoweb/1442).

PR #1445, opened 2025-04-24
Implement frontend for waitlist

brage-andreas:
image

### DIFF
--- a/apps/web/src/app/events/components/AttendanceCard/AttendanceCard.tsx
+++ b/apps/web/src/app/events/components/AttendanceCard/AttendanceCard.tsx
@@ -28,7 +28,7 @@ const getQueuePosition = (
   const index = attendees.filter((attendee) => !attendee.reserved).indexOf(attendee)
 
   if (index === -1) {
-    return undefined
+    return null
   }
 
   return index + 1
--- a/apps/web/src/app/events/components/AttendanceCard/AttendanceCard.tsx
+++ b/apps/web/src/app/events/components/AttendanceCard/AttendanceCard.tsx
@@ -1,7 +1,7 @@
 "use client"
 
 import { useTRPC } from "@/utils/trpc/client"
-import { type Attendance, type Attendee, type User, canUserAttendPool } from "@dotkomonline/types"
+import { type Attendance, type AttendancePool, type Attendee, type User, canUserAttendPool } from "@dotkomonline/types"
 import { Icon, Text, Title } from "@dotkomonline/ui"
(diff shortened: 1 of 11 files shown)


apps/web/src/app/events/components/AttendanceCard/AttendanceCard.tsx
```diff
@@ -15,13 +15,32 @@ import { TicketButton } from "./TicketButton"
 import { ViewAttendeesDialogButton } from "./ViewAttendeesButton"
 import { useDeregisterMutation, useRegisterMutation } from "./mutations"
 
+const getQueuePosition = (
+  attendee: Attendee | undefined,
+  attendees: Attendee[] | undefined,
+  attendablePool: AttendancePool | undefined
+) => {
```
junlarsen, line 22: It should probably not be possible to return undefined, null, or number from this. Maybe return `null` instead of undefined in the first branch

apps/web/src/app/events/components/AttendanceCard/RegistrationButton.tsx
```diff
@@ -43,7 +44,13 @@ export const RegistrationButton: FC<Props> = ({
...
   const className = clsx(
     "flex flex-row gap-2 items-center w-full text-black rounded-lg h-fit min-h-[4rem] p-2 text-left opacity-100 disabled:opacity-100",
-    disabled ? "bg-slate-4 text-slate-8" : attendee ? "bg-red-7 hover:bg-red-8" : "bg-green-7 hover:bg-green-8"
+    disabled
+      ? "bg-slate-4 text-slate-8"
+      : attendee
+        ? "bg-red-7 hover:bg-red-8"
+        : isPoolFull
+          ? "bg-yellow-5 hover:bg-yellow-6"
+          : "bg-green-7 hover:bg-green-8"
```
junlarsen, line 53: Triple nested ternary is a little suspicious but it's ok for now.

junlarsen (changes requested):
LGTM, just one small nit

apps/web/src/app/events/components/AttendanceCard/RegistrationButton.tsx
brage-andreas: I could make it a function 👍

junlarsen: I think creating functions for classnames is less ideal, because then the styles move quite far from the markup

PR #1447, opened 2025-04-24
Attempt to reserve next unreserved attendee on deregistration

brage-andreas:


junlarsen (approved):
nice! would writing a test for this be trivial? if not that's also OK

PR #1463, opened 2025-04-26
Migrate Company and Job Listings to Monoweb

sondrealf:
Unsure if we want to include the big queries in the PR.

But the manual queries to create it is included


junlarsen (changes requested):
This should probably not be a prisma migration, but instead just a mnaul insert script or something. We probably don't want to have this in the migration history

cc @henrikhorluck @henrikskog

junlarsen (approved):

PR #1482, opened 2025-04-30
Finalize job listing page styling so it looks good for lightmode

sondrealf:
Screenshot 2025-04-30 at 18 41 28

### DIFF
--- a/apps/web/src/components/views/JobListingView.tsx
+++ b/apps/web/src/components/views/JobListingView.tsx
@@ -98,26 +98,31 @@ export const JobListingView: FC<JobListingViewProps> = (props: JobListingViewPro
           <p className="m-0 text-4xl font-bold">{company.name}</p>
           <p className="m-0 text-3xl">{title}</p>
         </div>
-        <div
-          className="[&_[dir='ltr']]:text-white
-                      [&_[dir='ltr']]:border-amber-9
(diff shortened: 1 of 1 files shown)


apps/web/src/components/views/JobListingView.tsx
```diff
@@ -98,26 +98,31 @@ export const JobListingView: FC<JobListingViewProps> = (props: JobListingViewPro
...
-          />
+        <div className="ml-8">
+          <div className="border-amber-9 border-l-[1px] pl-4">
+            <div
+              className="[&_[dir='ltr']]:text-black
+                        [&_[dir='ltr']]:dark:text-white
+                        [&_ul]:list-disc
+                        [&_ol]:list-decimal
+                        [&_li]:ml-6
+                        mb-12 flex flex-col"
```
junlarsen, line 109: :thinking: Maybe we should just have tailwind prose or something..?

junlarsen (commented):
The yellow lines on the screenshot appear to be misaligned

apps/web/src/components/views/JobListingView.tsx
sondrealf: Tested, but then there will be much styling to overwrite, lineheight bullet point color etc

junlarsen (approved):

PR #1483, opened 2025-04-30
Script for migrating groups

henrikskog:
`cd tools/shell`
`doppler run --project monoweb-rpc --config dev -- tsx src/scripts/migrate-groups-from-ow4.ts`


andrefkl (approved):
works great, had to make the .json files manually but i guess its not that important for a once-sync

tools/shell/src/scripts/migrate-groups-from-ow4.ts
```diff
@@ -0,0 +1,216 @@
...
+    while (true) {
+      const response = await fetch(getHobbyUrl(page))
+      const data = await response.json()
+      if (data.next === null) {
+        break
+      }
+      page++
+      result.push(...data.results)
+
+      if (page > PAGE_LIMIT) {
```
junlarsen, line 55: Is PAGE_LIMIT defined in here?

tools/shell/src/scripts/migrate-groups-from-ow4.ts
```diff
@@ -0,0 +1,216 @@
...
+  other.forEach((item) => console.log(item.name_short))
+
+  console.log("\nINTEREST GROUPS: ")
+  // biome-ignore lint/complexity/noForEach: <explanation>
+  interestGroups.forEach((item) => console.log(item.title))
+
+  console.log("\n\nInserting interest groups:")
+  for (const item of interestGroups) {
+    console.log(`Inserting ${item.title}`)
+    await prisma.interestGroup.create({
```
junlarsen, line 103: Is there a reason we are doing direct prisma calls vs using the service layer? It's probably fine for these objects, but when it comes to things like events we have a lot of invariants we want to maintain

tools/shell/src/scripts/migrate-groups-from-ow4.ts
```diff
@@ -0,0 +1,216 @@
...
+program
+  .name("migrate-groups")
+  .description("CLI tool for migrating groups from OW4")
+  .addHelpText("beforeAll", () => {
+    printEnvironment()
+    return ""
+  })
+
+program
+  .command("1")
```
junlarsen, line 152: Maybe use more descriptive names than 1/2/3?

PR #1491, opened 2025-05-11
Add theme toggle in navbar

brage-andreas:
Unopened:
image

Opened:
image

### DIFF
--- a/packages/ui/src/atoms/Popover/Popover.tsx
+++ b/packages/ui/src/atoms/Popover/Popover.tsx
@@ -1,0 +1,28 @@
+"use client"
+
+import * as PopoverPrimitive from "@radix-ui/react-popover"
+import * as React from "react"
+import { cn } from "../../utils"
+
+export const Popover = PopoverPrimitive.Root
+
+export const PopoverTrigger = PopoverPrimitive.Trigger
+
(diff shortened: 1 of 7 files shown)


andrefkl:
Should this be used when the user is logged in too or do we still want that in the avatar dropdown?

packages/ui/src/atoms/Popover/Popover.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { cn } from "../../utils"
+
+export const Popover = PopoverPrimitive.Root
+
+export const PopoverTrigger = PopoverPrimitive.Trigger
+
+export const PopoverContent = React.forwardRef<
+  React.ElementRef<typeof PopoverPrimitive.Content>,
+  React.ComponentPropsWithoutRef<typeof PopoverPrimitive.Content>
+>(({ className, align = "center", sideOffset = 4, ...props }, ref) => (
```
junlarsen, line 14: Do not use forwardRef in React 19. Just take `ComponentPropsWithRef` and destructure ref from the props

junlarsen (changes requested):
Small nit on react code, otherwise LGTM

PR #1498, opened 2025-05-15
Reimplement selections

brage-andreas:
If you are missing some:

How the selection looks:

All done:


packages/core/src/modules/attendance/attendee-repository.ts
```diff
@@ -143,4 +145,35 @@ export class AttendeeRepositoryImpl implements AttendeeRepository {
...
+      },
+    })
+
+    const attendeesWithSelections = attendees.filter(
+      (attendee) =>
+        attendee.selections &&
+        typeof attendee.selections === "object" &&
+        Array.isArray(attendee.selections) &&
+        attendee.selections.length > 0
+    ) as { id: AttendeeId; selections: AttendanceSelectionResponse[] }[]
```
junlarsen, line 166: Same here with the conditions. Also, is the type assertion required? Does typing the function work?

```ts
.filter((thing): thing is { id: AttendeeId; selections: AttendanceSelectionResponse[] } => { ... }
```

junlarsen (changes requested):
Seems fine :shipit:. I think we can improve the readability on the long conditions however.

PR #1511, opened 2025-06-11
Create page for offlines

andrefkl:
image


brage-andreas (commented):
I think it could be nice to move the grouping/sorting into a separate method, but well done!

brage-andreas (approved):

junlarsen (approved):

PR #1512, opened 2025-06-14
Create article page

andrefkl:
https://github.com/user-attachments/assets/f7849ea1-396d-46c8-bcae-0a2fc710b0a1


apps/dashboard/src/app/(dashboard)/article/[id]/edit-card.tsx
```diff
@@ -10,7 +10,8 @@ export const ArticleEditCard: FC = () => {
   const FormComponent = useArticleWriteForm({
     label: "Oppdater artikkel",
     onSubmit: (data) => {
-      edit.mutate({ id: article.id, input: data })
+      const { tags, ...articleData } = data
+      edit.mutate({ id: article.id, article: articleData, tags })
```
Comment (junlarsen) [inline apps/dashboard/src/app/(dashboard)/article/[id]/edit-card.tsx:14]: maybe?
```suggestion
      const { tags, ...article } = data
      edit.mutate({ id: article.id, article, tags })
```

apps/dashboard/src/modules/article/use-article-table.tsx
```diff
@@ -1,7 +1,7 @@
 "use client"
 
 import type { Article } from "@dotkomonline/types"
-import { Anchor } from "@mantine/core"
+import { Anchor, Group as MantineGroup } from "@mantine/core"
```
junlarsen, line 4: Any reason we can't just use the import as Group?

packages/core/src/modules/article/article-service.ts
```diff
@@ -65,6 +81,40 @@ export class ArticleServiceImpl implements ArticleService {
...
+      .sort((a, b) => {
+        if (b.matchCount !== a.matchCount) {
+          return b.matchCount - a.matchCount
+        }
+
+        if (a.nonMatchCount !== b.nonMatchCount) {
+          return a.nonMatchCount - b.nonMatchCount
+        }
+
+        return b.article.updatedAt.getTime() - a.article.updatedAt.getTime()
```
junlarsen, line 112: you can probably return a compareAsc/compareDesc from date-fns here

junlarsen (approved):
LGTM - found some small nits but happy to accept as-is

PR #1514, opened 2025-06-17
Articles list page

andrefkl:
image


apps/web/src/components/molecules/ArticleFiltersContainer.tsx/index.tsx
```diff
@@ -0,0 +1,70 @@
+import type { ArticleTagName } from "@dotkomonline/types"
+import { Text, TextInput } from "@dotkomonline/ui"
+import clsx from "clsx"
+import Link from "next/link"
+import type { Dispatch, FC, SetStateAction } from "react"
+
+interface ArticleFiltersContainerProps {
+  search: string
+  setSearch: Dispatch<SetStateAction<string>>
```
junlarsen, line 9: This is probably good enough for now, but long-term we probably want to manage the search control state through a form state (react-hook-form).

PR #1515, opened 2025-06-18
Show featured articles on the front page

andrefkl:
image

### DIFF
--- a/apps/dashboard/src/app/(dashboard)/article/all-articles-table.tsx
+++ b/apps/dashboard/src/app/(dashboard)/article/all-articles-table.tsx
@@ -1,33 +1,37 @@
-"use client"
-
 import type { Article } from "@dotkomonline/types"
 import { Anchor, Group } from "@mantine/core"
-import { createColumnHelper, getCoreRowModel, useReactTable } from "@tanstack/react-table"
+import { createColumnHelper, getCoreRowModel } from "@tanstack/react-table"
 import Link from "next/link"
 import { useMemo } from "react"
 
-interface Props {
-  data: Article[]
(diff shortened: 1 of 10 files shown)


apps/dashboard/src/app/(dashboard)/article/all-articles-table.tsx
```diff
@@ -1,33 +1,37 @@
...
 import type { Article } from "@dotkomonline/types"
 import { Anchor, Group } from "@mantine/core"
-import { createColumnHelper, getCoreRowModel, useReactTable } from "@tanstack/react-table"
+import { createColumnHelper, getCoreRowModel } from "@tanstack/react-table"
 import Link from "next/link"
 import { useMemo } from "react"
 
-interface Props {
-  data: Article[]
+import { FilterableTable, arrayOrEqualsFilter } from "src/components/molecules/FilterableTable/FilterableTable"
```
junlarsen, line 7: nit: should use a relative path

packages/core/src/modules/article/article-service.ts
```diff
@@ -116,6 +117,10 @@ export class ArticleServiceImpl implements ArticleService {
       .slice(0, 10)
   }
 
+  async getFeatured(): Promise<Article[]> {
+    return (await this.articleRepository.getFeatured()).sort((a, b) => compareDesc(a.updatedAt, b.updatedAt))
```
junlarsen, line 121: nit: maybe move out the await expression
```suggestion
    const articles = await this.articleRepository.getFeatured()
    return articles.sort((a, b) => compareDesc(a.updatedAt, b.updatedAt))
```

junlarsen (changes requested):
Nice, excited to see the home page come together :sunglasses:

Spotted two very small nits, but otherwise LGTM

PR #1522, opened 2025-06-22
Invalidate dashboard queries

andrefkl:
Some attendance mutations needed eventId to invalidate, so they invalidate all queries instead to avoid passing eventId down to the mutation just for that.


brage-andreas (approved):
Nice!

PR #1562, opened 2025-07-06
Fix external package warnings

jotjern:


packages/core/package.json
```diff
@@ -34,6 +34,7 @@
     "@testcontainers/postgresql": "10.28.0",
     "@types/node": "22.15.29",
     "@vitest/ui": "3.1.4",
+    "jsdom": "26.1.0",
```
junlarsen, line 37: Ehhhhhhhhh

jsdom should never end up in the server bundle and is not used in core afaik

junlarsen (approved):

packages/core/package.json
jotjern: vitest depender

jotjern: kun dev dependency

PR #1570, opened 2025-07-06
Feedback forms

andrefkl:
Adds form to create feedback forms in dashboard and page in web to answer. Dashboard page to view and remove answers will be added in another pr. 

Forms and questions can be deleted if they don't have any connected answers. The form can only be answered after the event has ended. 

Not 100% sure which route to place the form under in web, think it makes most sense under [eventId], but there the event loading shows

image
image


apps/web/src/app/arrangementer/[slug]/[eventId]/tilbakemelding/page.tsx
```diff
@@ -0,0 +1,54 @@
...
+  if (!attendee) return <Text>Du kan ikke svare på dette skjemaet.</Text>
+
+  const previousAnswer = await server.event.feedback.findAnswerByAttendee.query({
+    formId: feedbackForm.id,
+    attendeeId: attendee.id,
+  })
+
+  if (previousAnswer) return <Text>Du har allerede svart på dette skjemaet.</Text>
+
+  if (event.end.getTime() > Date.now())
```
Comment (junlarsen) [inline apps/web/src/app/arrangementer/[slug]/[eventId]/tilbakemelding/page.tsx:31]: We should try to use date-fns as much as possible. Do you mind replacing with `isAfter(event.end, Date.now())`?

apps/dashboard/src/app/event/components/feedback-form-edit-form.tsx
```diff
@@ -0,0 +1,271 @@
...
+  type FeedbackFormId,
+  type FeedbackFormWrite,
+  FeedbackFormWriteSchema,
+  FeedbackQuestionSchema,
+  type FeedbackQuestionWrite,
+  FeedbackQuestionWriteSchema,
+} from "@dotkomonline/types"
+import { DragDropContext, Draggable, type DropResult, Droppable } from "@hello-pangea/dnd"
+import { zodResolver } from "@hookform/resolvers/zod"
+import { Icon } from "@iconify/react/dist/iconify.js"
```
junlarsen, line 11: I *think* we can just use the base import here

```suggestion
import { Icon } from "@iconify/react"
```

junlarsen (changes requested):
Awesome, nice work! I have two small nits, please let me know what you think 🚀

PR #1583, opened 2025-07-08
Execute tasks on JobExecutor concurrently

junlarsen:
NOTE: Concurrently does of course not mean in parallel - we are working in Node after all, and we're not using worker_threads.

Instead of maintaining a running state for the job executor, we should instead maintain running states on each individual job. Instead of blocking on all executions, we can now fire off jobs as soon as their seen, and thanks to database consistency, we will (hopefully) not see the same job twice.

The execution of the task has been changed to use its own Prisma transaction for isolation, while keeping the task state changes outside of said transaction.


junlarsen:
* **#1583** <a href="https://app.graphite.dev/github/pr/dotkom/monoweb/1583?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/dotkom/monoweb/1583?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

brage-andreas (approved):

PR #1663, opened 2025-07-29
Implement marks

jotjern:
## Implemented features:

- [x] Create mark manually
- [x] Create suspension manually
- [x] Implement dashboard page for giving out marks
- [x] Add profile page section for marks
- [x] Handle holidays
- [x] Add penalty for having marks

## Not yet implemented:
* Send out email to groups hosting events if users have missed attendance and should be given penalties.
* Add dashboard buttons on events for giving out marks to users who missed attendance
* Send out emails when marks are handed out
* Show information about delayed event registration on event page


jotjern:
* **#1663** <a href="https://app.graphite.dev/github/pr/dotkom/monoweb/1663?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/dotkom/monoweb/1663?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/rpc/src/invariant.ts
```diff
@@ -13,7 +13,7 @@ const logger = getLogger("db-query-invariant-checker")
 export function parseOrReport<T>(schema: z.ZodSchema<T>, value: T): T {
   const result = schema.safeParse(value)
   if (!result.success) {
-    logger.error("Database failed to parse value into schema: %s emitted for object %o", result.error.message, value)
+    logger.error("Database failed to parse value into schema: %s emitted for object %o", result, value)
```
junlarsen, line 16: Needs %o for the formatter to format it nicely
```suggestion
    logger.error("Database failed to parse value into schema: %o emitted for object %o", result, value)
```

packages/db/prisma/migrations/20250729174946_marks_refactor/migration.sql
```diff
@@ -0,0 +1,26 @@
...
+
+*/
+-- CreateEnum
+CREATE TYPE "MarkType" AS ENUM ('MANUAL', 'LATE_ATTENDANCE', 'MISSED_ATTENDANCE');
+
+-- AlterTable
+ALTER TABLE "mark" DROP COLUMN "category",
+ADD COLUMN     "groupSlug" TEXT NOT NULL,
+ADD COLUMN     "type" "MarkType" NOT NULL DEFAULT 'MANUAL',
+ADD COLUMN     "weight" INTEGER NOT NULL;
```
junlarsen, line 16: We're probably OK since nor staging or production have values here, but for future reference:

`groupSlug` and `weight` here will cause the database migration to fail if there are rows in the table, as they are marked `NOT NULL` and have no default to populate them.

To fix this, the generated SQL for the migration should be modified

```sql
-- Create a dummy group to hold all the missing values
INSERT INTO "group" (slug, ...) VALUES ('migration-placeholder', ...);

-- Populate the existing columns with either dummy values, or a sensible default
ALTER TABLE "mark" DROP COLUMN "category",
ADD COLUMN "groupSlug" TEXT NOT NULL DEFAULT (SELECT "slug" FROM "group" WHERE ...),
ALTER COLUMN "groupSlug" DROP DEFAULT,
ADD COLUMN "weight" INTEGER NOT NULL DEFAULT 1,
ALTER COLUMN "weight" DROP DEFAULT;
```

Then you would later move over values manually from the dummy group to realistic ones (in our domain, set them to HS probably), then create a new migration with the contents
```sql
DELETE FROM "group" WHERE "slug" = 'migration-placeholder';
```

Again, should be fine to not do it for this PR, but important to keep in mind for future reference.

apps/rpc/src/invariant.ts
jotjern: Nice thanks

apps/rpc/src/modules/mark/personal-mark-repository.ts
```diff
@@ -29,12 +31,46 @@ export function getPersonalMarkRepository(): PersonalMarkRepository {
       const personalMarks = await handle.personalMark.findMany({ where: { userId }, select: { mark: true } })
       return personalMarks.map((personalMark) => parseOrReport(MarkSchema, personalMark.mark))
     },
+    async getDashboardByMarkId(handle, markId) {
```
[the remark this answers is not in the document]
jotjern: Because the dashboard needs different stuff from the web app. I agree in principal, but they need to be different endpoints as I don't think the user should get to know who specifically gave them the mark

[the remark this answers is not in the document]
jotjern: Also, the dashboard needs user information, which is not present on the personalmark schema as of now. The schemas are not set in stone though, I might change them around

junlarsen (changes requested):
Nice work - I found some small nits with regards to naming etc. Let me know what you think

junlarsen (approved):

PR #1693, opened 2025-07-31
Update events in dashboard

brage-andreas:
This PR:
- Updates event list in dashboard
  - Should note that while a search bar is added, it is disabled and non-functional
- Adds a delete button to events

image

image

### DIFF
--- a/apps/dashboard/src/app/event/page.tsx
+++ b/apps/dashboard/src/app/event/page.tsx
@@ -1,16 +1,50 @@
 "use client"
 
-import type { Event } from "@dotkomonline/types"
-import { formatDate } from "@dotkomonline/utils"
+import type { Event, EventType } from "@dotkomonline/types"
 import { Icon } from "@iconify/react"
-import { Anchor, Box, Button, ButtonGroup, Skeleton, Stack } from "@mantine/core"
+import {
+  ActionIcon,
+  ActionIconGroup,
+  Anchor,
+  Button,
+  Group,
+  Skeleton,
+  Stack,
+  Text,
+  TextInput,
+  Title,
+  Tooltip,
+} from "@mantine/core"
 import { createColumnHelper, getCoreRowModel, useReactTable } from "@tanstack/react-table"
 import Link from "next/link"
 import { useMemo } from "react"
 
 import { GenericTable } from "@/components/GenericTable"
+import { formatDate } from "date-fns"
+import { nb } from "date-fns/locale"
 import { EventHostingGroupList } from "./components/event-hosting-group-list"
 import { useEventAllQuery } from "./queries"
+
+const mapEventTypeToLabel = (eventType: EventType) => {
+  switch (eventType) {
+    case "ACADEMIC":
+      return "Kurs"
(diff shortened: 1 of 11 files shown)


apps/dashboard/src/app/event/page.tsx
```diff
@@ -1,17 +1,51 @@
...
+  Title,
+  Tooltip,
+} from "@mantine/core"
 import { createColumnHelper, getCoreRowModel, useReactTable } from "@tanstack/react-table"
 import Link from "next/link"
 import { useMemo } from "react"
 
 import { GenericTable } from "@/components/GenericTable"
+import { formatDate } from "date-fns"
+import { nb } from "date-fns/locale"
```
junlarsen, line 24: We should set locale options for date-fns in the root layout.tsx instead

apps/dashboard/src/app/event/page.tsx
```diff
@@ -53,20 +99,32 @@ export default function EventPage() {
...
+            <ActionIconGroup>
+              <ActionIcon variant="subtle" c="gray" size="input-sm">
+                <Icon icon="tabler:arrow-left" />
+              </ActionIcon>
+              <ActionIcon variant="subtle" c="gray" size="input-sm">
+                <Icon icon="tabler:arrow-right" />
+              </ActionIcon>
+            </ActionIconGroup>
+
+            {/* TODO: add search */}
```
junlarsen, line 116: Not to sound all agile coachy, but is there an issue on Linear for this?

apps/rpc/src/modules/event/event-repository.ts
```diff
@@ -56,13 +56,13 @@ export function getEventRepository(): EventRepository {
   return {
     async create(handle, data) {
       const row = await handle.event.create({ data })
-      const event = await this.findById(handle, row.id)
+      const event = await this.findById(handle, row.id, { includeDeleted: true })
```
junlarsen, line 59: Shouldn't this be false? Under what circumstances would an event be deleted at the same time of creation?

apps/dashboard/src/app/event/page.tsx
brage-andreas: Can do this in another PR

packages/types/src/event.ts
```diff
@@ -22,10 +22,11 @@ export const EventSchema = schemas.EventSchema.extend({
 
 export const EventTypeSchema = schemas.EventTypeSchema
 export const EventStatusSchema = schemas.EventStatusSchema
+export const EventStatusWriteSchema = schemas.EventStatusSchema.exclude(["DELETED"])
+export type EventStatusWrite = z.infer<typeof EventStatusWriteSchema>
```
junlarsen, line 26: What is this used for?

brage-andreas: Creating drop downs for status in edit and create event form

junlarsen: Hmm okay, although I would argue if its not used elsewhere, just declare that in the dashboard code

PR #1703, opened 2025-08-02
Fix membership calculation and authentication logic

junlarsen:


junlarsen:
* **#1703** <a href="https://app.graphite.dev/github/pr/dotkom/monoweb/1703?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/dotkom/monoweb/1703?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/rpc/src/modules/user/user-service.ts
```diff
@@ -152,44 +161,77 @@ export function getUserService(
     async findUsers(handle, query, page) {
       return await userRepository.findMany(handle, query, page ?? { take: 20 })
     },
-    async register(handle, subject) {
-      const accessToken = await this.findFeideAccessTokenByUserId(subject)
-      const user = await userRepository.register(handle, subject)
+    async register(handle, userId) {
```
junlarsen, line 164: Most of the logic in the PR is inside this function. We:

1. Do nothing if there is no Feide connection to create and find membership for
2. Never update the user data based on what we get after the user has been created. Before this, you would get a new profile slug when you added a Feide connection.
3. Runs the membership refresh logic inside an OpenTelemetry span so its easier to trace what is going on.

brage-andreas (approved):

PR #1735, opened 2025-08-04
Implement punishment delay in frontend

brage-andreas:
Delay from marks:

Delay from marks (dark mode):

Suspension from marks:

Suspension from marks (dark mode):

After you register:

Closes MONO-511

### DIFF
--- a/apps/rpc/src/modules/mark/personal-mark-service.ts
+++ b/apps/rpc/src/modules/mark/personal-mark-service.ts
@@ -8,7 +8,7 @@ import type {
   UserId,
   VisiblePersonalMarkDetails,
 } from "@dotkomonline/types"
-import { getExpiryDate, unique } from "@dotkomonline/utils"
+import { getPunishmentExpiryDate, unique } from "@dotkomonline/utils"
 import { isPast } from "date-fns"
 import type { GroupService } from "../group/group-service"
 import type { MarkService } from "./mark-service"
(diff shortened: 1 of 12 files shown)


apps/rpc/src/modules/mark/personal-mark-service.ts
```diff
@@ -98,25 +98,25 @@ export function getPersonalMarkService(
...
         if (!mark) return acc
-        if (isPast(getExpiryDate(pm.createdAt, mark.duration))) return acc
+        if (isPast(getPunishmentExpiryDate(pm.createdAt, mark.duration))) return acc
 
         return acc + mark.weight
       }, 0)
 
       if (currentMarkWeight >= 6) {
-        return { suspended: true }
+        return { delay: 0, suspended: true }
```
junlarsen, line 107: I would prefer the delay here to be in milliseconds. There is also no documentation on what unit this is

apps/web/src/app/arrangementer/components/AttendanceCard/PunishmentBox.tsx
```diff
@@ -0,0 +1,60 @@
+import type { Punishment } from "@dotkomonline/types"
+import { Button, Icon, Text } from "@dotkomonline/ui"
+
+export const PunishmentBox = ({ punishment }: { punishment: Punishment }) => {
```
junlarsen, line 4: Lets be consistent with declaring components. Most other places in the codebase uses FC

packages/db/prisma/schema.prisma
```diff
@@ -305,6 +305,7 @@ model Mark {
   id       String  @id @default(uuid())
   title    String
   details  String?
+  /// Duration in days
```
junlarsen, line 308: Should be milliseconds as bove

packages/types/src/mark.ts
```diff
@@ -42,8 +42,9 @@ export const PersonalMarkDetailsSchema = z.object({
 })
 
 export const PunishmentSchema = z.object({
-  suspended: z.boolean().optional(),
-  delay: z.number().optional(),
+  suspended: z.boolean(),
+  /** Delay in hours */
```
junlarsen, line 46: Same here

packages/utils/src/holidays.ts
```diff
@@ -1,33 +1,24 @@
...
+
+const JANUARY = 0 as const
+const JUNE = 5 as const
+const AUGUST = 7 as const
+const DECEMBER = 11 as const
 
 function getHolidaysThisYear(): Interval[] {
-  const yearToday = new Date().getFullYear()
+  const currentYear = getYear(new Date())
+  const nextYear = getYear(addYears(new Date(), 1))
```
junlarsen, line 11: Replace `new Date()` with getCurrentUtc()

junlarsen (changes requested):

apps/rpc/src/modules/mark/personal-mark-service.ts
junlarsen: Might as well

PR #1736, opened 2025-08-04
Filter and page articles serverside

andrefkl:
Same design with some color fixes, changed to richtext viewers and moved filtering to rpc. Switched to only showing top 30 tags cause there were way too many to show them all


apps/rpc/src/query.ts
```diff
@@ -1,12 +1,11 @@
...
-export const PaginateInputSchema = z
-  .object({
-    take: z.number(),
-    cursor: z.string().uuid().optional(),
-  })
-  .optional()
-  .default({ take: 20, cursor: undefined })
+export const BasePaginateInputSchema = z.object({
+  take: z.number().default(20),
+  cursor: z.string().uuid().optional(),
```
junlarsen, line 5: Do you mind changing cursor to not be a uuid? we have tables whose primary key isn't a uuid

apps/rpc/src/modules/article/article-repository.ts
```diff
@@ -18,7 +20,9 @@ export interface ArticleRepository {
   getById(handle: DBHandle, articleId: ArticleId): Promise<Article | null>
   getBySlug(handle: DBHandle, slug: ArticleSlug): Promise<Article | null>
   getByTags(handle: DBHandle, tags: ArticleTagName[], page?: Pageable): Promise<Article[]>
+  findMany(handle: DBHandle, query: ArticleFilterQuery, page: Pageable): Promise<Article[]>
   getFeatured(handle: DBHandle): Promise<Article[]>
+  getMostUsedTags(handle: DBHandle, take: number): Promise<ArticleTag[]>
```
junlarsen, line 25: nit: rename to findTagsOrderedByPopularity

junlarsen (approved):
Awesome

PR #1737, opened 2025-08-04
Fix authorization for personal marks

brage-andreas:
This PR:
- Adds `requireMe` and `requireMeOrAffiliated` authorization helpers to trpc ctx
- Tightens all routes for `personalMark`


apps/rpc/src/trpc.ts
```diff
@@ -44,6 +44,28 @@ export const createContext = async (principal: Principal | null, context: Servic
...
+        require(principal.subject === userId)
+      },
+      /**
+       * Requires either `requireMe` or `requireAffiliation` to be true.
+       *
+       * One of the following must be true:
+       * - The user is signed in and that the provided user id is the user's id.
+       * - The user is a member of at least one of the provided groups (or any group if empty).
+       */
+      requireMeOrAffiliation({ userId, affiliations }: { userId: UserId; affiliations: Affiliation[] }) {
```
junlarsen, line 62: I would prefer to pass two arguments instead here. It's rather easy to distinguish a scalar from an array value

junlarsen (changes requested):

PR #1752, opened 2025-08-06
Add infinite scroll and search to event list

andrefkl:


apps/rpc/src/modules/event/event-router.ts
```diff
@@ -89,28 +83,21 @@ export const eventRouter = t.router({
...
-        })
-        .optional()
-    )
-    .output(z.array(EventSchema))
+    .input(BasePaginateInputSchema.extend({ filter: EventFilterQuerySchema.optional() }).optional())
+    .query(async ({ input, ctx }) =>
+      ctx.executeTransaction(async (handle) => await ctx.eventService.findEvents(handle, { ...input?.filter }, input))
+    ),
+
+  allPaged: procedure
```
junlarsen, line 91: I'm a little confused, why do we need both all and allPages? These should be one query

packages/types/src/event.ts
```diff
@@ -43,8 +43,10 @@ export const EventFilterQuerySchema = z
   .object({
     byId: buildAnyOfFilter(EventSchema.shape.id),
     byStartDate: buildDateRangeFilter(),
+    byEndDate: buildDateRangeFilter(),
     bySearchTerm: buildSearchFilter(),
     byOrganizingCompany: buildAnyOfFilter(CompanySchema.shape.id),
     byOrganizingGroup: buildAnyOfFilter(GroupSchema.shape.slug),
+    orderBy: createSortOrder(),
```
junlarsen, line 50: Later we should allow selecting what field(s) to sort, but this is fine for now

junlarsen (changes requested):

PR #1759, opened 2025-08-08
Payments

jotjern:
This PR aims to implement payment for events via stripe.

Will refactor some tomorrow, but would appreciate some feedback. Here is a quick explanation:

In stripe you have:
* Products - Each event is a product
* Price - You create a price for a product to be able to pay. In this implementation we only have one price at a time. 
* Checkout session - Each attendee's payment is a checkout session
* Payment intent - Created when an attendee pays

This implementation works by:
When an `Attendance` is updated with a value for `attendancePrice`, the `paymentService.createOrUpdateProduct` is called, which makes a stripe product with a description, name, price etc matching the event and attendance.

Whenever an attendee reserves a spot in an event that has a attendanceprice, `paymentService.createProductPayment` is called, and some fields on `Attendee` are populated: `paymentDeadline`, `paymentLink` and `paymentId`. A stripe checkout session is created, and the payment link is available to the user on the event attendance card. 

When the users completes the payment, a webhook is triggered that sets `paymentReservedAt` (or `paymentChargedAt` if the de-registration deadline has passed)

If the de-registration deadline has not passed, the payment is only reserved, and a task called `CHARGE_ATTENDANCE_PAYMENTS` (`apps/rpc/src/modules/task/task-definition.ts`) runs when the de-registration deadline is. This makes sure users are instantly refunded if they decide to deregister. 

When attendees move off the waitlist they get a 24 hour payment limit, but when they instantly get a spot they only have 15 minutes. When they do a `VERIFY_PAYMENT` task is started, and after the deadline they are kicked off the event if they have not yet paid. 

TODO:
- [x] Add a scheduled task to execute charges for events
- [x] Charges should be immediate if timeout is reached
- [x] Add refund button to dashboard
- [x] Add payment information to dashboard
- [x] Use one stripe account/key
- [x] Add metadata to stripe product from event
- [x] When a user has been charged, show a graphic of why they cannot unnattend

### DIFF
--- a/apps/rpc/src/modules/attendance/attendee-service.ts
+++ b/apps/rpc/src/modules/attendance/attendee-service.ts
@@ -18,7 +18,8 @@ import {
   getMembershipGrade,
 } from "@dotkomonline/types"
 import { getCurrentUTC } from "@dotkomonline/utils"
-import { addHours, addMinutes, addSeconds, isFuture, isPast } from "date-fns"
+import { captureException } from "@sentry/node"
+import { addHours, addMinutes, isFuture, isPast } from "date-fns"
 import type { PersonalMarkService } from "../mark/personal-mark-service"
 import { PaymentAlreadyChargedError, PaymentUnexpectedStateError } from "../payment/payment-error"
 import type { Payment, PaymentService } from "../payment/payment-service"
@@ -36,7 +37,6 @@ import {
   AttendeeRegistrationError,
 } from "./attendee-error"
 import type { AttendeeRepository } from "./attendee-repository"
-import { captureException } from "@sentry/node"
 
 type AdminDeregisterForEventOptions = { reserveNextAttendee: boolean; bypassCriteriaOnReserveNextAttendee: boolean }
 


apps/rpc/src/bin/server.ts
```diff
@@ -79,6 +81,34 @@ server.get("/health", (_, res) => {
   res.send({ status: "ok" })
 })
 
+server.post("/webhook/stripe", async (req, res) => {
+  console.log(req.body)
```
junlarsen, line 85:
```suggestion
```

apps/rpc/src/modules/attendance/attendee-service.ts
```diff
@@ -254,26 +347,146 @@ export function getAttendeeService(
...
+      if (!attendance.attendancePrice) {
+        throw new Error("Tried to create a payment for an attendance without a price")
+      }
+
+      const payment = await paymentService.create(
+        attendance.id,
+        isPast(attendance.deregisterDeadline) ? "CHARGE" : "RESERVE"
+      )
+
+      taskSchedulingService.scheduleAt(
```
junlarsen, line 380: Missing await

packages/utils/src/ogjoin.ts
```diff
@@ -0,0 +1,8 @@
+export const ogJoin = (names: string[]) => {
```
junlarsen, line 1: 👀

junlarsen (commented):
generally lgtm

PR #1841, opened 2025-08-11
Fix event list regressions and group events

brage-andreas:


apps/rpc/src/modules/event/event-router.ts
```diff
@@ -19,14 +20,31 @@ export const eventRouter = t.router({
...
-      ctx.executeTransaction(async (handle) => ctx.eventService.findEventById(handle, input))
+      ctx.executeTransaction(async (handle) => {
+        const event = await ctx.eventService.findEventById(handle, input)
+        if (!event) {
+          return null
+        }
+        const attendance = event.attendanceId
+          ? await ctx.attendanceService.findAttendanceById(handle, event.attendanceId)
+          : null
+        return { ...event, attendance }
```
junlarsen, line 46: Maybe return both objects instead of spreading the event?

apps/rpc/src/modules/event/event-router.ts
```diff
@@ -37,15 +55,17 @@ export const eventRouter = t.router({
...
-        return await ctx.eventService.updateEventOrganizers(
+        const eventWithoutOrganizers = await ctx.eventService.createEvent(handle, input.event)
+        const event = await ctx.eventService.updateEventOrganizers(
           handle,
-          event.id,
+          eventWithoutOrganizers.id,
           new Set(input.groupIds),
           new Set(input.companies)
         )
+        return { ...event, attendance: null }
```
junlarsen, line 68: Same here and everywhere else too

apps/web/src/app/arrangementer/components/EventList.tsx
```diff
@@ -2,68 +2,122 @@
...
+  futureEvents,
+  pastEvents,
+  fetchNextPastEventsPage,
+}: EventListProps) => {
   const session = useSession()
+  const now = getCurrentUTC()
 
-  const futureEventItems = futureEvents.map((eventDetail) => mapEventDetailToItem(eventDetail))
-  const pastEventItems = pastEvents.map((eventDetail) => mapEventDetailToItem(eventDetail))
+  const [yourEvents, openEvents, openingSoonEvents, otherFutureEvents] = futureEvents.reduce<
```
junlarsen, line 25: I would much rather reduce into an object

apps/dashboard/src/app/company/[slug]/company-page.tsx
```diff
@@ -1,5 +1,5 @@
 import { GenericTable } from "@/components/GenericTable"
-import type { Event } from "@dotkomonline/types"
+import type { EventDetail } from "@dotkomonline/types"
```
Comment (junlarsen) [inline apps/dashboard/src/app/company/[slug]/company-page.tsx:2]: Prefer to rename this to EventWithAttendance probably, EventDetail is not very descriptive

apps/web/src/app/arrangementer/components/AttendanceCard/AttendanceCard.tsx
```diff
@@ -32,9 +32,9 @@ import { TicketButton } from "./TicketButton"
 import { ViewAttendeesButton } from "./ViewAttendeesButton"
 
 const getQueuePosition = (
-  attendee: Attendee | undefined,
-  attendees: Attendee[] | undefined,
-  attendablePool: AttendancePool | undefined
+  attendee: Attendee | null | undefined,
+  attendees: Attendee[] | null | undefined,
+  attendablePool: AttendancePool | null | undefined
```
junlarsen, line 37: Is this actually a tri-state or not? Under what circumstances can the attendance be:

1. There
2. Null
3. Undefined

It sounds like it should be either T|null or T|undefined, not T|undefined|null

junlarsen (changes requested):

junlarsen (approved):

PR #1871, opened 2025-08-13
Hide past events in Splash

brage-andreas:


apps/splash-2025/package.json
```diff
@@ -17,6 +17,7 @@
     "@dotkomonline/utils": "workspace:*",
     "@fontsource/figtree": "^5.2.8",
     "@fontsource/inter": "^5.2.6",
+    "@formkit/auto-animate": "^0.8.2",
```
junlarsen, line 20: I don't particularly like adding new dependencies for a single thing. Couldn't we also just do this with regular css transforms/transitions? I'm not sure I see why we need a new dependency for this

junlarsen (changes requested):

PR #1884, opened 2025-08-15
Correct event colors in list view and calendar view

henrygraesberg:


apps/web/src/components/organisms/EventCalendar/index.tsx
```diff
@@ -51,19 +51,19 @@ function getColSpanClass(span: number) {
 
 function getEventTypeGuide(events: Event[]) {
   const eventTypeConfig = {
-    SOCIAL: { color: "bg-green-400 dark:bg-green-400", label: "Sosialt" },
-    ACADEMIC: { color: "bg-red-400 dark:bg-red-400", label: "Kurs" },
+    SOCIAL: { color: "bg-red-400 dark:bg-red-400", label: "Sosialt" },
```
junlarsen, line 54: Social should be green, company should be red

junlarsen (changes requested):

apps/web/src/components/organisms/EventCalendar/index.tsx
henrygraesberg: In that case, list view is wrong as well - Did check OW4 and social was green and company red there, so I'll fix it in both views

junlarsen (approved):

PR #1892, opened 2025-08-15
Add SSE and live countdown to attendance

Terbau:
- Now sends new attendees (both when registrating and deregistrating) to all connected clients. This means that attendees are now automatically updated on the client without having to refresh the site. 
- Adds a countdown to the event card and automatically renders the registration button as green when the registration has opened.

Nice quality of life updates

image

### DIFF
--- a/.github/actions/build-brevduen/action.yml
+++ b/.github/actions/build-brevduen/action.yml
@@ -14,7 +14,7 @@ runs:
   using: composite
   steps:
     - name: Install local GitHub Actions
-      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4
       with:
         fetch-depth: 1
         sparse-checkout: '.github'
--- a/.github/actions/build-dashboard/action.yml
+++ b/.github/actions/build-dashboard/action.yml
@@ -14,7 +14,7 @@ runs:
   using: composite
   steps:
     - name: Install local GitHub Actions
-      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4
       with:
         fetch-depth: 1
         sparse-checkout: '.github'
--- a/.github/actions/build-invoicification/action.yml
+++ b/.github/actions/build-invoicification/action.yml
@@ -14,7 +14,7 @@ runs:
   using: composite
   steps:
     - name: Install local GitHub Actions
-      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4
       with:
         fetch-depth: 1
         sparse-checkout: '.github'
--- a/.github/actions/build-kvittering-backend/action.yml
+++ b/.github/actions/build-kvittering-backend/action.yml
@@ -11,7 +11,7 @@ runs:
   using: composite
   steps:
     - name: Install local GitHub Actions
-      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4
       with:
         fetch-depth: 1
         sparse-checkout: '.github'
--- a/.github/actions/build-kvittering-frontend/action.yml
+++ b/.github/actions/build-kvittering-frontend/action.yml
@@ -11,7 +11,7 @@ runs:
   using: composite
   steps:
     - name: Install local GitHub Actions
-      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4
       with:
         fetch-depth: 1
         sparse-checkout: '.github'
@@ -22,7 +22,7 @@ runs:
         project: monoweb-kvittering-frontend
         config: prd
 
-    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+    - uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4
       with:
         fetch-depth: 1
         ref: ${{ inputs.git-hash }}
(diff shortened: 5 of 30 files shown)


apps/rpc/src/modules/event/attendance-router.ts
```diff
@@ -136,6 +138,15 @@ export const attendanceRouter = t.router({
       })
     ),
 
+  onRegisterChange: procedure
+    .input(z.object({ attendanceId: AttendanceSchema.shape.id }))
+    .subscription(async function* ({ ctx, signal }) {
```
junlarsen, line 143: 🗣️ async 🗣️ generator 🗣️

apps/web/src/utils/trpc/QueryProvider.tsx
```diff
@@ -4,42 +4,81 @@ import { env } from "@/env"
...
+
+export interface TRPCSSERegisterChangeConnectionStateContextType {
+  trpcSSERegisterChangeConnectionState: TRPCSSEConnectionState
+  setTRPCSSERegisterChangeConnectionState: Dispatch<SetStateAction<TRPCSSEConnectionState>>
+}
+
+const TRPCSSERegisterChangeConnectionStateContext =
+  createContext<TRPCSSERegisterChangeConnectionStateContextType | null>(null)
+
+export const useTRPCSSERegisterChangeConnectionState = () => {
```
junlarsen, line 31: name of the century

junlarsen (approved):
lgtm, needs some linting though

apps/rpc/src/modules/event/attendance-router.ts
```diff
@@ -136,6 +138,15 @@ export const attendanceRouter = t.router({
...
+      for await (const [data] of on(ctx.eventEmitter, "attendance:register-change", { signal })) {
+        const attendeeUpdateData = data as { attendee: Attendee; status: "registered" | "deregistered" }
+        yield attendeeUpdateData
+      }
+    }),
+
```
brage-andreas, line 149: This might need some documentation or references, since I have no idea why it is a generator like this or how to make another one

apps/rpc/src/trpc.ts
```diff
@@ -80,6 +81,16 @@ export const t = initTRPC.context<Context>().create({
...
     return shape
   },
+  sse: {
+    maxDurationMs: minutesToMilliseconds(20), // 20 minutes
+    ping: {
+      enabled: true,
+      intervalMs: secondsToMilliseconds(3), // 30 seconds
+    },
+    client: {
+      reconnectAfterInactivityMs: secondsToMilliseconds(5), // 5 seconds
```
brage-andreas, line 91:
```suggestion
    maxDurationMs: minutesToMilliseconds(20),
    ping: {
      enabled: true,
      intervalMs: secondsToMilliseconds(3),
    },
    client: {
      reconnectAfterInactivityMs: secondsToMilliseconds(5),
```

apps/web/src/app/arrangementer/components/AttendanceCard/AttendanceCard.tsx
```diff
@@ -77,6 +83,49 @@ export const AttendanceCard = ({ user, initialAttendance, initialPunishment }: A
...
+              return {
+                ...oldData,
+                attendees: [...oldData.attendees, attendeeUpdateData.attendee],
+              }
+            }
+          )
+        },
+      }
+    )
+  )
```
brage-andreas, line 127: amen🙏🕊️☠️🪦

apps/web/src/app/arrangementer/components/mutations.ts
```diff
@@ -15,11 +17,13 @@ export const useDeregisterMutation = () => {
           return
         }
 
-        await Promise.all([
-          await queryClient.invalidateQueries(
-            trpc.event.attendance.getAttendance.queryOptions({ id: input.attendanceId })
-          ),
-        ])
+        if (trpcSSERegisterChangeConnectionState !== "pending") {
```
brage-andreas, line 20: Maybe a comment saying that `pending` means the connection is alive, and that we only invalidate since the connection can't provide the fresh data? Otherwise I think this could confuse people as to why it is here

apps/web/src/app/arrangementer/components/mutations.ts
```diff
@@ -32,13 +36,16 @@ interface UseRegisterMutationInput {
...
   const queryClient = useQueryClient()
+  const { trpcSSERegisterChangeConnectionState } = useTRPCSSERegisterChangeConnectionState()
 
   return useMutation(
     trpc.event.attendance.registerForEvent.mutationOptions({
       onSuccess: async (data) => {
-        await Promise.all([
-          queryClient.invalidateQueries(trpc.event.attendance.getAttendance.queryOptions({ id: data.attendanceId })),
-        ])
+        if (trpcSSERegisterChangeConnectionState !== "pending") {
```
brage-andreas, line 44: same here as in deregister mutation

apps/web/src/utils/trpc/QueryProvider.tsx
```diff
@@ -4,42 +4,81 @@ import { env } from "@/env"
...
       }),
     ],
   }
   const [queryClient] = useState(
     () =>
       new QueryClient({
         defaultOptions: {
+          queries: {
+            staleTime: secondsToMilliseconds(30),
+          },
```
brage-andreas, line 81: Is this for all queries? Does this make every single query refetch every 30 seconds?

apps/web/src/utils/utils.ts
brage-andreas, line 1: In the future this component should probably sync itself to the nearest second

brage-andreas (approved):

apps/rpc/src/modules/event/attendance-router.ts
brage-andreas: Maybe it should also validate the data it gets and throw if its invalid

junlarsen: generators are state machines that can yield multiple times

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncGenerator

apps/web/src/utils/trpc/QueryProvider.tsx
junlarsen: No, by default the staleTime is zero. This means that each subscriber to a query will refetch it (after deduplication) as data is considered stale by default.

This change means it's going not going to refetch more often than every 30 seconds, unless the query is invalidated by us

https://tanstack.com/query/v4/docs/framework/react/guides/important-defaults

apps/web/src/utils/utils.ts
junlarsen: setInterval/setTimeout are not precise, the best we can do is to calculate based on the time delta. I think this is finee

junlarsen (approved):

PR #1912, opened 2025-08-16
Respect draft events

brage-andreas:


junlarsen (changes requested):

packages/types/src/event.ts
```diff
@@ -49,5 +49,6 @@ export const EventFilterQuerySchema = z
     byOrganizingGroup: buildAnyOfFilter(GroupSchema.shape.slug),
     excludingOrganizingGroup: buildAnyOfFilter(GroupSchema.shape.slug),
     orderBy: createSortOrder(),
+    includeStatuses: z.array(EventStatusSchema.exclude(["PUBLIC"])),
```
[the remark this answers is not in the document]
junlarsen: I also don't know why we have excludingOrganizingGroup? It does the same as byOrganizingGroup..?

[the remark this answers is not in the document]
junlarsen: yeah I suppose it's fine

apps/dashboard/src/app/event/queries.ts
```diff
@@ -13,7 +13,7 @@ interface UseEventAllQueryProps {
 export const useEventAllQuery = ({ filter, page }: UseEventAllQueryProps) => {
   const trpc = useTRPC()
   const { data, ...query } = useQuery({
-    ...trpc.event.all.queryOptions({ filter: { ...filter, includeStatuses: ["DRAFT"] }, ...page }),
+    ...trpc.event.all.queryOptions({ filter: { ...filter, includeStatus: ["DRAFT"] }, ...page }),
```
junlarsen, line 16: Should be named `byStatus`

brage-andreas: Wouldn't that be misleading?

brage-andreas: Or do you want to default to `["PUBLIC"]` and if you give just `["DRAFT"]` you get only drafy

junlarsen: Probably yeah, we should probably make some consistent rules for filter naming but I'd like to keep it byStatus to be somewhat consistent until we make further decisions

PR #1941, opened 2025-08-20
Add parent event to event object

junlarsen:
Disallows circular graphs and height>2 for now.


apps/rpc/src/modules/event/attendance-service.ts
```diff
@@ -282,7 +282,6 @@ export function getAttendanceService(
     async registerAttendee(handle, attendanceId, userId, options) {
       const attendance = await this.getAttendanceById(handle, attendanceId)
       const user = await userService.getById(handle, userId)
-      const event = await eventService.getByAttendance(handle, attendance.id)
```
junlarsen, line 285: There is a ton of broken tests due to a regression from default find rules @brage-andreas. Adding this line again and running tests breaks 12 test cases.

PR #1955, opened 2025-08-20
Allow specific affiliations

brage-andreas:


junlarsen (changes requested):
I'm not sure I understand the changes, it doesn't appear to bring any new functionality from what I can tell

junlarsen:
Hmm, I would honestly just prefer we do it in the procedure body. That way you also catch them in the middleware etc

PR #1963, opened 2025-08-21
Add backend check for attendee registration in parent events

brage-andreas:


andrefkl (approved):

apps/rpc/src/modules/event/attendance-service.ts
```diff
@@ -300,6 +303,35 @@ export function getAttendanceService(
         )
       }
 
+      if (event.parentId) {
+        if (options.ignoreRegisteredToParent) {
+          logger.info(
```
junlarsen, line 308: Isnt there intended to be more code here?

brage-andreas: bruh

brage-andreas: if it ignores it just logs and moves on

PR #1965, opened 2025-08-21
Add copy feedback form option

andrefkl:
image


apps/rpc/src/modules/event/event-repository.ts
```diff
@@ -133,6 +133,17 @@ export function getEventRepository(): EventRepository {
                   }
                 : undefined,
             },
+            {
```
junlarsen, line 136: monster query brewing (not a problem, we moght need indexes though, problem for later 🍺)

junlarsen (approved):
nice, definitely going to be useful 🍻🍻🍻

PR #2004, opened 2025-08-24
Fix 404 on settings if not signed in

brage-andreas:


andrefkl:
Maybe /profil should do the same thing instead of redirecting to home?

packages/utils/src/create-authorize-url.ts
```diff
@@ -0,0 +1,9 @@
+export const createAuthorizeUrl = (...params: { [name: string]: string }[]) => {
+  const url = new URL("/api/auth/authorize", window.location.origin)
+  for (const param of params) {
+    for (const [name, value] of Object.entries(param)) {
+      url.searchParams.set(name, value)
+    }
+  }
+  return url.toString()
+}
```
junlarsen, line 9: Is there a reason there's an array type + spread?

Also, could we instead just pass URLSearchParameter's constructor parameters?


```suggestion
export const createAuthorizeUrl = (...parameters: ConstructorParameters<URLSearchParams>) => {
  const url = new URL("/api/auth/authorize", window.location.origin)
  url.search = new URLSearchParams(...parameters).toString()
  return url.toString()
}
```

PR #2007, opened 2025-08-25
Improve event searching UX

brage-andreas:
Adds a delete search button, prevents submission (so enter doesn't clear the text), and moves skeleton to just table

### DIFF
--- a/apps/dashboard/src/app/(internal)/event/components/event-filters.tsx
+++ b/apps/dashboard/src/app/(internal)/event/components/event-filters.tsx
@@ -1,6 +1,7 @@
 import type { EventFilterQuery } from "@dotkomonline/types"
-import { TextInput } from "@mantine/core"
+import { ActionIcon, Group, TextInput } from "@mantine/core"
 import { useDebouncedValue } from "@mantine/hooks"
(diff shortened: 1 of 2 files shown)


apps/dashboard/src/app/(internal)/event/components/event-filters.tsx
```diff
@@ -18,8 +19,19 @@ export const EventFilters = ({ onChange }: Props) => {
...
-      <TextInput placeholder="Søk etter arrangementer..." {...form.register("bySearchTerm")} />
+    <form
+      onSubmit={(e) => {
+        e.preventDefault()
+      }}
+    >
+      <Group gap={4}>
+        <TextInput placeholder="Søk etter arrangementer..." {...form.register("bySearchTerm")} />
+        {Boolean(data.bySearchTerm) && (
+          <ActionIcon size="input-sm" variant="subtle" color="gray" type="reset" onClick={() => form.reset()}>
```
junlarsen, line 30: Should probably be single field reset and not the whole form? Make it correct now so we don't have to fix it when we add more filters

junlarsen (changes requested):

PR #2008, opened 2025-08-26
Make QR code scanning more consistent

brage-andreas:
Various changes, like increasing video feed size, QR-code size and error correction level, and making sure you can't scan more QR codes while a QR-code modal is open

### DIFF
--- a/apps/dashboard/src/app/(internal)/event/components/qr-code-scanned-modal.tsx
+++ b/apps/dashboard/src/app/(internal)/event/components/qr-code-scanned-modal.tsx
@@ -29,6 +29,7 @@ export const QRCodeScannedModal: FC<ContextModalProps<ModalProps>> = ({
   id,
   innerProps: { attendance, attendeeId },
 }) => {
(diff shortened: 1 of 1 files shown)


apps/dashboard/src/app/(internal)/event/components/qr-code-scanner.tsx
```diff
@@ -14,6 +14,7 @@ interface QrCodeScannerProps {
 export const QrCodeScanner: FC<QrCodeScannerProps> = ({ attendance }) => {
   const [scannerOpen, { toggle: toggleScanner }] = useDisclosure(false)
   const [videoReady, setVideoReady] = useState(false)
+  const paused = useRef(false)
```
junlarsen, line 17: Is there a particular reason this is a ref and not a state?

brage-andreas: Yes. Calling useState from the modal gives an error, since a component is trying to update another component

junlarsen (approved):

BillySBarrett (approved):

PR #2035, opened 2025-08-30
Move payment countdown to main card

brage-andreas:
image

image

image


apps/rpc/src/modules/event/attendance-service.ts
```diff
@@ -137,6 +137,7 @@ export interface AttendanceService {
   updateAttendeeById(handle: DBHandle, attendeeId: AttendeeId, data: Partial<AttendeeWrite>): Promise<Attendee>
   executeReserveAttendeeTask(handle: DBHandle, task: InferTaskData<ReserveAttendeeTaskDefinition>): Promise<void>
   deregisterAttendee(handle: DBHandle, attendeeId: AttendeeId, options: EventDeregistrationOptions): Promise<void>
+  findChargeAttendeeScheduleDate(handle: DBHandle, attendeeId: AttendeeId): Promise<Date | null>
```
junlarsen, line 140: This should probably be just be stored on attendance instead of having to query the task system.

I also suspect this function to be really slow because we have no indexing on any of the JSON column fields

brage-andreas:
> I don't know if I agree with sorting the dates. It's a little confusing to see the dates hop around

It's a timeline so i think it will be more confusing if it doesn't sort. It highlights with a yellow background if it changes so I don't think it would be a problem

apps/rpc/src/modules/event/attendance-router.ts
```diff
@@ -276,4 +276,21 @@ export const attendanceRouter = t.router({
...
+  findChargeAttendeeScheduleDate: authenticatedProcedure
+    .input(z.object({ attendeeId: AttendeeSchema.shape.id }))
+    .output(z.date().nullable())
+    .query(async ({ input, ctx }) => {
+      return ctx.executeTransaction(async (handle) => {
+        const attendee = await ctx.attendanceService.getAttendeeById(handle, input.attendeeId)
+
+        if (!attendee) {
+          return null
+        }
```
andrefkl, line 289: Isn't this never undefined?

apps/web/src/app/arrangementer/components/AttendanceCard/MainPoolCard.tsx
```diff
@@ -10,236 +10,481 @@ import {
...
+  const paymentCountdownInterval =
+    attendee?.createdAt && attendee.paymentDeadline ? interval(attendee.createdAt, attendee.paymentDeadline) : null
+  const isWithinPaymentCountdown = paymentCountdownInterval ? isWithinInterval(now, paymentCountdownInterval) : false
+
+  const hasPaid = Boolean(
+    attendee &&
+      (attendee.paymentChargedAt ||
+        attendee.paymentReservedAt ||
+        (attendee.paymentRefundedAt && !attendee.paymentDeadline))
+  )
```
andrefkl, line 62: Maybe this should be moved to a function since it's used 3 times

andrefkl (approved):
Just some nits but lgtm!

PR #2041, opened 2025-09-01
Hide internal events from non-staff

brage-andreas:


junlarsen (commented):
As far as I understand this only hides them on the webpage. Maybe that's enough?

In other words, do you think we want to hide them on the server too perhaps?

junlarsen:
> Do you mean to only serve internal events if the request is done by a staff? Is that better?

Yeah, but I'm not sure if it's worth the effort. I don't think there's particularly private or secret stuff. I'm mostly wondering what you think

brage-andreas:
> > Do you mean to only serve internal events if the request is done by a staff? Is that better?
> 
> Yeah, but I'm not sure if it's worth the effort. I don't think there's particularly private or secret stuff. I'm mostly wondering what you think

Hmm idk. I can make the exludingType default to internal, so you can explicitly pass an empty array to include them

junlarsen (approved):

PR #2048, opened 2025-09-03
Fix logout link prefetch causing logout

andrefkl:
Opening the profile dropdown logs the user out in the background because of link prefetch


andrefkl:
Yeah hahaha

PR #2051, opened 2025-09-03
Add Google Workspace sync

brage-andreas:
- [x] Rotate service account in Google Cloud Console
- [x] Add `workspace-service.ts`
- [x] Add `workspace-router.ts`
- [x] It works
- [ ] Add UI in dashboard
- [ ] Link `User#workspaceUserId` with any existing workspace user accounts

Closes MONO-504

### DIFF
--- a/apps/rpc/src/modules/workspace-sync/workspace-service.ts
+++ b/apps/rpc/src/modules/workspace-sync/workspace-service.ts
@@ -4,7 +4,7 @@ import { getLogger } from "@dotkomonline/logger"
 import type { Group, GroupId, GroupMember, User, UserId } from "@dotkomonline/types"
 import { slugify } from "@dotkomonline/utils"
 import type { admin_directory_v1 } from "googleapis"
-import type { GaxiosResponseWithHTTP2 } from "googleapis-common"
+import { GaxiosError, type GaxiosResponseWithHTTP2 } from "googleapis-common"
 import invariant from "tiny-invariant"
 import { configuration } from "../../configuration"
 import type { GroupService } from "../group/group-service"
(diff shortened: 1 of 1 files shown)


apps/rpc/src/modules/workspace-sync/client.ts
```diff
@@ -0,0 +1,58 @@
...
+  client_id: z.string().min(1).max(100),
+  client_x509_cert_url: z.string().url(),
+  private_key: z.string().min(1).max(1000),
+  private_key_id: z.string().min(1).max(100),
+  project_id: z.string().min(1).max(100),
+  token_uri: z.string().url(),
+  type: z.literal("service_account"),
+})
+
+const getJWT = () => {
```
junlarsen, line 26: Pretty sure you can authenticate against google apis using just the service account, and not forging a jwt

apps/rpc/src/modules/workspace-sync/helpers.ts
```diff
@@ -0,0 +1,60 @@
+import { randomBytes } from "node:crypto"
+import type { Group, User } from "@dotkomonline/types"
+import { slugify } from "@dotkomonline/utils"
+
+const DEFAULT_DOMAIN = process.env.WORKSPACE_SYNC_DOMAIN ?? "online.ntnu.no"
```
junlarsen, line 5: Should be defined in configuration.ts

apps/rpc/src/modules/workspace-sync/users.ts
```diff
@@ -0,0 +1,117 @@
...
+      recoveryPhone: user.phone,
+      changePasswordAtNextLogin: true,
+    },
+  })
+
+  const is2faEnforced = response.data.isEnforcedIn2Sv ?? false
+  const is2faEnabled = response.data.isEnrolledIn2Sv ?? false
+  let recoveryCodes: string[] | null = null
+
+  if (is2faEnforced && !is2faEnabled) {
```
junlarsen, line 76: Why do we have to generate mfa recovery coees for the user? Shouldnt they do it themselves?

junlarsen (changes requested):

PR #2052, opened 2025-09-04
Show feedback forms missing answer in profile page

andrefkl:


apps/rpc/src/modules/event/event-repository.ts
```diff
@@ -58,6 +58,7 @@ export interface EventRepository {
   deleteEventCompanies(handle: DBHandle, eventId: EventId, companyIds: Set<CompanyId>): Promise<void>
   updateEventAttendance(handle: DBHandle, eventId: EventId, attendanceId: AttendanceId): Promise<Event>
   updateEventParent(handle: DBHandle, eventId: EventId, parentEventId: EventId | null): Promise<Event>
+  findByUserNotGivenFeedback(handle: DBHandle, userId: UserId): Promise<Event[]>
```
junlarsen, line 61: Maybe name it findUnansweredByUser?

PR #2085, opened 2025-09-07
Allow non-staff to fetch their own feedback answers

andrefkl:


apps/rpc/src/modules/event/feedback-router.ts
```diff
@@ -106,6 +107,23 @@ export const feedbackRouter = t.router({
...
+        formId: FeedbackFormIdSchema,
+        attendeeId: AttendeeSchema.shape.id,
+      })
+    )
+    .query(async ({ input, ctx }) =>
+      ctx.executeTransaction(async (handle) => {
+        const attendee = await ctx.attendanceRepository.findAttendeeById(handle, input.attendeeId)
+        if (attendee?.userId !== ctx.principal.subject) {
+          throw new TRPCError({ code: "UNAUTHORIZED", message: "Unauthorized" })
+        }
```
brage-andreas, line 122:
```suggestion
        ctx.authorize.requireMe(attendee.userId)
```

PR #2086, opened 2025-09-08
Suspend user if missing payment past deadline

junlarsen:


brage-andreas (approved):

apps/rpc/src/modules/event/attendance-service.ts
```diff
@@ -833,7 +834,30 @@ export function getAttendanceService(
...
+          handle,
+          {
+            details: `Suspensjon for å ikke betale for arrangement ${event.title}`,
+            // We do not have a method for indefinite duration yet.
+            duration: 100_000,
+            title: "Suspensjon for mangelende betaling",
+            type: "MISSING_PAYMENT",
+            // Immediate suspension
+            weight: 6,
+            userIds: [attendee.userId],
```
andrefkl, line 849: A bit late but i don't think userIds actually does anything and that personalMarkService also needs to be called (idk why its on the MarkWrite type)

PR #2100, opened 2025-09-08
Add recurring tasks

andrefkl:
Suggestion for how we can do it, lmk what you think.
Before each task discovery, pending recurring tasks are used to schedule normal tasks so we preserve history.

### DIFF
--- a/packages/types/src/task.ts
+++ b/packages/types/src/task.ts
@@ -2,17 +2,33 @@ import { schemas } from "@dotkomonline/db/schemas"
 import type { z } from "zod"
 
 export const TaskSchema = schemas.TaskSchema.extend({})
+export const RecurringTaskSchema = schemas.RecurringTaskSchema.extend({})
 
 export type Task = z.infer<typeof TaskSchema>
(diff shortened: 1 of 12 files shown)


packages/types/src/task.ts
```diff
@@ -2,17 +2,33 @@ import { schemas } from "@dotkomonline/db/schemas"
...
 
 export type TaskWrite = z.infer<typeof TaskWriteSchema>
+
+export const RecurringTaskWriteSchema = RecurringTaskSchema.pick({
+  type: true,
+  schedule: true,
+  payload: true,
+  lastRunAt: true,
+}).partial({
+  lastRunAt: true,
```
brage-andreas, line 31:
```suggestion
}).partial({
  lastRunAt: true,
```

brage-andreas (approved):

packages/types/src/task.ts
andrefkl: It doesn't like that 🤷

junlarsen (approved):

PR #2105, opened 2025-09-09
Require dependencies to be 14 days old

junlarsen:
This should prevent us from eating most bad packages (like what happened with debug+friends yesterday)


andrefkl (approved):
Niiice

PR #2109, opened 2025-09-09
Move feedback form ping to correct link when profile menu is open

brage-andreas:
image
image

### DIFF
--- a/apps/web/src/components/Navbar/ProfileMenu.tsx
+++ b/apps/web/src/components/Navbar/ProfileMenu.tsx
@@ -17,89 +17,13 @@ import {
   DropdownMenuSeparator,
   DropdownMenuTrigger,
   Icon,
+  Text,
 } from "@dotkomonline/ui"
 import { createAuthorizeUrl, createLogoutUrl } from "@dotkomonline/utils"
-import { skipToken, useQuery } from "@tanstack/react-query"
+import { skipToken, useQueries, useQuery } from "@tanstack/react-query"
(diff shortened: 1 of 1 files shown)


apps/web/src/components/Navbar/ProfileMenu.tsx
```diff
@@ -184,18 +119,84 @@ export const AvatarDropdown: FC<PropsWithChildren> = ({ children }) => {
...
+              </div>
+            </DropdownMenuContent>
+          </DropdownMenu>
+        </div>
+      </div>
+    )
+  }
+
+  const showFeedbackFormPing = eventsMissingFeedback && eventsMissingFeedback.length > 0
+  const showPing = showFeedbackFormPing
```
andrefkl, line 170:
```suggestion
  const showFeedbackFormPing = eventsMissingFeedback && eventsMissingFeedback.length > 0
```

andrefkl (approved):
Just the one nit but otherwise lgtm

apps/web/src/components/Navbar/ProfileMenu.tsx
brage-andreas: Sure. I meant this to make it less buggy if some other ping is added in the future. Like say we also want pings for new marks and suspensions

andrefkl: Oooh ok i get it nvm then

PR #2119, opened 2025-09-10
Small fixes for-bedrifter

Virham:
Text in image stays black when in dark mode, fixed typo and animation on lines


johannes-aas:
looks good

apps/web/src/app/for-bedrifter/index.css
johannes-aas, line 1: maybe call this dashAnimation.css instead of index

PR #2174, opened 2025-09-16
Fix migration

brage-andreas:


junlarsen (changes requested):
Dont modify existing migration files. Create a new one

junlarsen:
Replaced by https://github.com/dotkom/monoweb/pull/2175

PR #2180, opened 2025-09-16
Create barebones email submission service

junlarsen:
Based on the same definition registry system that the task serviceu ses.

### DIFF
--- a/apps/rpc/src/modules/user/user-service.ts
+++ b/apps/rpc/src/modules/user/user-service.ts
@@ -226,33 +226,35 @@ export function getUserService(
       if (accessToken !== null) {
         // We spawn a separate OpenTelemetry span for the entire membership operation so that its easier to trace and
         // track the call stack and timings of the operation.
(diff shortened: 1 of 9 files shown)


apps/rpc/src/modules/user/user-service.ts
```diff
@@ -226,33 +226,35 @@ export function getUserService(
...
-              // We can only replace memberships if there is a new applicable one for the user
-              if (
-                shouldReplaceMembership(user.memberships, activeMembership, applicableMembership) &&
-                applicableMembership !== null
-              ) {
-                logger.info("Discovered applicable membership for user %s: %o", user.id, applicableMembership)
-                await userRepository.createMembership(handle, user.id, applicableMembership)
+        await trace
+          .getTracer("@dotkomonline/rpc/user-service")
+          .startActiveSpan("UserService#refreshMembership", async (span) => {
```
junlarsen, line 231: The only diff here is the span name

PR #2189, opened 2025-09-17
Enable richtext editing for group description

jennicaduong:


andrefkl (changes requested):
Nice! I think we need to change this to use the `RichText` component in web so it renders correctly too
https://github.com/dotkom/monoweb/blob/c257cc401721c13136242969f0db6925be167788/apps/web/src/app/komiteer/components/GroupPage.tsx#L95

PR #2206, opened 2025-09-18
Add audit log

ErlendSae:
## Created auditlog

Tracks every insert, update and delete happening in the database. 
Shows:
1. **who** did the action
2. **when** the action happened
3.  **what** type of operation
4. **what** they performed the action on

image

## Details page
After clicking "se detaljer" you get all the changes done, as well as link to relevant table (doesnt work for joined tables), fex an event that has been changed.
| image | image |
|--------|--------|

### On inserts and deletes it looks like this:
image

## Improvements/missing
-  I've tracked the transactionID for each transaction making it better to just display each transaction rather than every entry in the database in the future
- Filtering based on type++
- Github-like diff - don't know how feasable, as only the diff gets stored in updates (but at least something better than current frontend)

### DIFF
--- a/apps/dashboard/src/app/ApplicationShell.tsx
+++ b/apps/dashboard/src/app/ApplicationShell.tsx
@@ -31,7 +31,7 @@ import {
 } from "@tabler/icons-react"
 import Link from "next/link"
 import { usePathname } from "next/navigation"
-import { type FC, type PropsWithChildren, useEffect } from "react"
+import { type FC, useEffect } from "react"
 
 const navigations = [
   {
@@ -84,15 +84,21 @@ const navigations = [
     label: "Hendelseslogg",
     icon: IconClipboardList,
     href: "/audit-log",
+    isAdmin: true,
   },
 ] satisfies {
   label: string
   icon: FC
   href: string
   openInNewTab?: boolean
+  isAdmin?: boolean
 }[]
+interface ApplicationShellProps {
+  isAdmin: boolean
+  children: React.ReactNode
+}
 
-export const ApplicationShell: FC<PropsWithChildren> = ({ children }) => {
+export const ApplicationShell: FC<ApplicationShellProps> = ({ isAdmin, children }) => {
   const [mobileOpened, { toggle: toggleMobile }] = useDisclosure()
   const [desktopOpened, { toggle: toggleDesktop }] = useDisclosure(true)
   const pathname = usePathname()
@@ -134,19 +140,21 @@ export const ApplicationShell: FC<PropsWithChildren> = ({ children }) => {
         </Group>
       </AppShellHeader>
       <AppShellNavbar p="md">
-        {navigations.map((navigation) => (
-          <NavLink
-            component={Link}
-            key={navigation.label}
-            label={navigation.label}
-            href={navigation.href}
-            active={pathname.startsWith(navigation.href)}
-            variant="subtle"
-            leftSection={<navigation.icon width={18} height={18} />}
-            style={{ borderRadius: "var(--mantine-radius-md)" }}
-            {...(navigation.openInNewTab ? { target: "_blank", rel: "noopener noreferrer" } : {})}
-          />
-        ))}
+        {navigations
+          .filter((navigation) => isAdmin || !navigation.isAdmin)
+          .map((navigation) => (
+            <NavLink
+              component={Link}
+              key={navigation.label}
+              label={navigation.label}
+              href={navigation.href}
+              active={pathname.startsWith(navigation.href)}
+              variant="subtle"
+              leftSection={<navigation.icon width={18} height={18} />}
+              style={{ borderRadius: "var(--mantine-radius-md)" }}
+              {...(navigation.openInNewTab ? { target: "_blank", rel: "noopener noreferrer" } : {})}
+            />
+          ))}
         <Button component="a" variant="outline" href="/api/auth/logout" hiddenFrom="xs" mt="lg">
           Logg ut
         </Button>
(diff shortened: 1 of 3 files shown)


apps/rpc/src/modules/core.ts
```diff
@@ -132,6 +134,20 @@ export async function createServiceLayer(
   clients: ReturnType<typeof createThirdPartyClients>,
   configuration: Configuration
 ) {
+  async function executeTransactionWithAudit<T>(
```
junlarsen, line 137: nit: rename this to executeAuditedTransaction maybe?

apps/rpc/src/modules/core.ts
```diff
@@ -132,6 +134,20 @@ export async function createServiceLayer(
...
+    userId: string | null
+  ): Promise<T> {
+    return clients.prisma.$transaction(async (tx) => {
+      const transactionId = crypto.randomUUID()
+
+      await tx.$executeRaw`SELECT set_config('app.current_user_id', ${userId || null}, TRUE)`
+      await tx.$executeRaw`SELECT set_config('app.current_transaction_id', ${transactionId}, TRUE)`
+
+      return fn(tx as typeof clients.prisma)
+    })
```
junlarsen, line 148: PostgreSQL has a transaction ID for every transaction. We should just use that instead.

Also, this cast is a little suspicious. Perhaps you should see if these types are sound (aka no assertions)


```suggestion
    fn: (tx: DBHandle) => Promise<T>,
    userId: UserId| null
  ): Promise<T> {
    return clients.prisma.$transaction(async (tx) => {
      await tx.$executeRaw`SELECT set_config('app.current_user_id', ${userId || null}, TRUE)`
      await tx.$executeRaw`SELECT set_config('app.current_transaction_id', 'pg_current_xact_id()', TRUE)`

      return fn(tx)
    })
```

apps/rpc/src/trpc.ts
```diff
@@ -37,6 +37,11 @@ export const createContext = async (principal: Principal | null, context: Servic
   return {
     ...context,
     principal,
+
+    executeTransactionWithAudit<T>(fn: (tx: typeof context.prisma) => Promise<T>): Promise<T> {
```
junlarsen, line 41: Adjust types here if the above suggestion works fine

packages/db/prisma/migrations/20250911143328_add_audit_log/migration.sql
```diff
@@ -0,0 +1,16 @@
+-- CreateTable
+CREATE TABLE "audit_log" (
+    "id" TEXT NOT NULL,
+    "tableName" TEXT NOT NULL,
+    "rowId" TEXT,
+    "createdAt" TIMESTAMPTZ(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "userId" TEXT,
+    "operation" TEXT NOT NULL,
+    "rowData" JSONB NOT NULL,
+    "transactionId" TEXT,
```
junlarsen, line 10: This would have to be of database type `xid8` in Postgres

packages/db/prisma/migrations/20250911143357_audit_log_trigger_function/migration.sql
```diff
@@ -0,0 +1,173 @@
...
+              FROM (
+                SELECT o.key, o.value AS old_val, n.value AS new_val
+                FROM jsonb_each_text(to_jsonb(OLD)) AS o(key, value)
+                JOIN jsonb_each_text(to_jsonb(NEW)) AS n(key, value) USING (key)
+                WHERE o.value IS DISTINCT FROM n.value
+              ) diffs
+            )
+            ELSE NULL
+        END),
+        current_setting('app.current_transaction_id', true)
```
junlarsen, line 61: Alternatively you could just drop setting the `app.current_transaction_id` value, and use pg_current_xact_id() directly here

packages/db/prisma/schema.prisma
```diff
@@ -653,3 +653,19 @@ model FeedbackFormAnswer {
...
+  tableName   String
+  rowId       String?
+  createdAt   DateTime @default(now()) @db.Timestamptz(3)
+  user        User?    @relation(fields: [userId], references: [id])
+  userId      String?
+  operation   String
+  rowData     Json
+
+  // Adds transaction context
+  transactionId String?
```
junlarsen, line 668: I think `BigInt` should work here. Maybe we have to do `pg_current_xact_id()::bigint` in the trigger

junlarsen (changes requested):
I would much rather prefer if we could use PostgreSQLs own transaction ids than creating our own.

I also think there are some type improvements to be made.

apps/rpc/src/trpc.ts
junlarsen: I also wonder if we can just write ```
executeAuditedTransaction: context.executeAuditedTransaction.bind(context),
```

apps/rpc/src/modules/core.ts
ErlendSae: yep

packages/db/prisma/migrations/20250911143328_add_audit_log/migration.sql
ErlendSae: Yep casted to bigint

packages/db/prisma/migrations/20250911143357_audit_log_trigger_function/migration.sql
ErlendSae: Did drop it, works well

junlarsen (approved):
Awesome - great work

brage-andreas (approved):

PR #2223, opened 2025-09-20
Send email with feedback form link day after event

andrefkl:
Will run once a day at 12 using a recurring task

image


apps/rpc/src/modules/email/email-template.ts
```diff
@@ -86,5 +91,18 @@ export const emails = {
       }),
     getTemplate: async () => fsp.readFile(path.join(templates, "company_invoice_notification.mustache"), "utf-8"),
   }),
+  FEEDBACK_FORM_LINK: createEmailTemplate({
+    type: "FEEDBACK_FORM_LINK",
+    getSchema: () =>
+      z.object({
+        eventName: z.string(),
```
junlarsen, line 98: Matbe just take the Event object?

apps/rpc/src/modules/event/attendance-service.ts
```diff
@@ -941,6 +958,66 @@ export function getAttendanceService(
...
+    async executeSendFeedbackFormLinkEmails(handle) {
+      const eventsEndedYesterday = await eventService.findEvents(handle, {
+        byHasFeedbackForm: true,
+        byEndDate: {
+          min: new TZDate(startOfYesterday()),
+          max: new TZDate(endOfYesterday()),
+        },
+      })
+
+      for (const event of eventsEndedYesterday) {
```
junlarsen, line 970: SES fulfillment is quite slow, gather all send promises and Promise.all them

apps/rpc/src/modules/email/email-template.ts
junlarsen: Hmm right this is fine for now then :+1:

PR #2240, opened 2025-09-24
Change default joblisting dates to today

andrefkl:
Bedkom finds it pretty annoying to always change the year when adding a joblisting


henrygraesberg (approved):
LGTM!

PR #2279, opened 2025-09-28
Move registration availability into separate service function

junlarsen:
Moves all logical checks for attendance requirements into separate function. In the long term, the client should use this function instead of calculating all of the values on its own. Enables a single source of truth for attendance checks.

Also fixed up integration tests so they pass

### DIFF
--- a/apps/rpc/src/modules/event/attendance-service.ts
+++ b/apps/rpc/src/modules/event/attendance-service.ts
@@ -1,7 +1,7 @@
 import type { EventEmitter } from "node:events"
 import { TZDate } from "@date-fns/tz"
 import type { DBHandle } from "@dotkomonline/db"
-import { getLogger } from "@dotkomonline/logger"
+import { type Logger, getLogger } from "@dotkomonline/logger"
 import {
   type Attendance,
   type AttendanceId,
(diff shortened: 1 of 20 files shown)


junlarsen:
The same should be done for deregistration

apps/rpc/src/modules/event/attendance-service.ts
```diff
@@ -1,7 +1,7 @@
 import type { EventEmitter } from "node:events"
 import { TZDate } from "@date-fns/tz"
-import type { DBHandle } from "@dotkomonline/db"
-import { getLogger } from "@dotkomonline/logger"
+import type { DBHandle, Membership } from "@dotkomonline/db"
```
andrefkl, line 3: Import Membership from /types

andrefkl (approved):

apps/rpc/src/modules/event/attendance-service.ts
junlarsen: oh yup lol, we should honestly have some guardrails here, same with `src/foo` imports of which I know we have a lot of

andrefkl: yeah truue

PR #2340, opened 2025-10-09
Remove `give-me-staff` script

brage-andreas:
Removes `give-me-staff` script and instead checks if the instance is running on a local environment. If and only if it is a developer environment, all `requireAffiliation` and `requireMeOrAffiliation` checks will succeed.

### DIFF
--- a/apps/rpc/src/configuration.ts
+++ b/apps/rpc/src/configuration.ts
@@ -88,6 +88,4 @@ export function isAmazonSesEmailFeatureEnabled(
 // We do not set NODE_ENV to development in our containers, so we just check that it's not staging or production
 // This should be safe enough with Dopplers own environment as well.
 export const isDevelopmentEnvironment =
-  process.env.DOPPLER_ENVIRONMENT === "dev" &&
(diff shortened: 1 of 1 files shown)


apps/rpc/src/configuration.ts
```diff
@@ -84,3 +84,10 @@ export function isAmazonSesEmailFeatureEnabled(
 ): configuration is ConfigurationWithAmazonSesEmail {
   return configuration.email.awsSqsQueueUrl !== null
 }
+
+// We do not set NODE_ENV to development in our containers, so we just check that it's not staging or production
+// This should be safe enough with Dopplers own environment as well.
+export const isDevelopmentEnvironment =
+  process.env.DOPPLER_ENVIRONMENT === "dev" &&
+  process.env.NODE_ENV !== "development" &&
+  process.env.NODE_ENV !== "staging"
```
junlarsen:
```suggestion
  process.env.NODE_ENV !== "production" &&
  process.env.NODE_ENV !== "staging"
```

PR #2353, opened 2025-10-13
Add workspace sync frontend

brage-andreas:
## Grupper
image
image
image
image
image
image
image

## Brukere
image
image
image

## Ikke admin
image
image
image

### DIFF
--- a/apps/rpc/src/modules/workspace-sync/workspace-service.ts
+++ b/apps/rpc/src/modules/workspace-sync/workspace-service.ts
@@ -367,8 +367,8 @@ export function getWorkspaceService(
       return workspaceUser
     },
 
-    async getWorkspaceUser(handle, userId) {
-      const workspaceUser = await this.findWorkspaceUser(handle, userId)
+    async getWorkspaceUser(handle, userId, customKey) {
+      const workspaceUser = await this.findWorkspaceUser(handle, userId, customKey)
 
       if (!workspaceUser) {
         throw new NotFoundError(`Workspace User for User(ID=${userId}) not found`)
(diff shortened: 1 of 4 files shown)


packages/types/src/workspace-sync.ts
```diff
@@ -0,0 +1,14 @@
...
+
+// Re-export types from googleapis package
+export const WorkspaceUserSchema = z.custom<admin_directory_v1.Schema$User>()
+export const WorkspaceGroupSchema = z.custom<admin_directory_v1.Schema$Group>()
+export const WorkspaceMemberSchema = z.custom<admin_directory_v1.Schema$Member>()
+export type WorkspaceUser = admin_directory_v1.Schema$User
+export type WorkspaceGroup = admin_directory_v1.Schema$Group
+export type WorkspaceMember = admin_directory_v1.Schema$Member
+
+export const WorkspaceMemberSyncActionSchema = z.enum(["NONE", "NEEDS_LINKING", "TO_ADD", "TO_REMOVE"])
```
junlarsen: Does this need to be a zod schema/sent over the wire?

apps/rpc/src/modules/workspace-sync/workspace-service.ts
```diff
@@ -151,6 +141,16 @@ export function getWorkspaceService(
...
   }
 
+  async function removeFromWorkspaceGroup(groupKey: string, memberKey: string): Promise<boolean> {
+    return await directory.members
+      .delete({
+        groupKey,
+        memberKey,
+      })
+      .then(() => true)
+      .catch(() => false)
```
junlarsen: code style: use try catch instead

apps/rpc/src/modules/workspace-sync/workspace-service.ts
```diff
@@ -167,6 +167,22 @@ export function getWorkspaceService(
     return `${local}@${domain}`
   }
 
+  function getLinkedWorkspaceId(localResolvable: User | Group | string): string | null {
```
junlarsen: I think we should find a more descriptive name than localResolvable

junlarsen: At this point a type alias would also be suitable. This union is repeated quite a lot

apps/rpc/src/modules/workspace-sync/workspace-service.ts
brage-andreas: why?

PR #2356, opened 2025-10-15
Add anniversary as event type

henrygraesberg:


junlarsen (commented):
Hm sure, but do you think adding another enum kind is the best solution - or what is the rationale for doing so?

PR #2373, opened 2025-10-19
Add modal for collecting deregistration reason

andrefkl:
For event types academic and company:
image

For any other event types:
image

### DIFF
--- a/packages/types/src/attendance.ts
+++ b/packages/types/src/attendance.ts
@@ -112,12 +112,6 @@ export const DeregisterReasonWriteSchema = DeregisterReasonSchema.pick({
   userGrade: true,
 })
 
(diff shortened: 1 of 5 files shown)


packages/types/src/attendance.ts
```diff
@@ -99,6 +99,49 @@ export const AttendanceWriteSchema = AttendanceSchema.pick({
...
+  eventId: true,
+  registeredAt: true,
+  userGrade: true,
+})
+
+export type DeregisterReasonFormWrite = z.infer<typeof DeregisterReasonFormWriteSchema>
+export const DeregisterReasonFormWriteSchema = DeregisterReasonWriteSchema.pick({
+  type: true,
+  details: true,
+})
```
junlarsen: we havent shared form types from server/client before. maybe just duplicate the types for now?

PR #2379, opened 2025-10-22
Fix kvittering error handling to display errors to users

henrikskog:
Previously, errors from PDF generation and email sending were only sent to Slack error channel without showing details to users. Now:

- Extract actual error messages from backend API responses
- Display detailed error messages in toast notifications
- Include error details in Slack alerts for debugging
- Improve error logging with Sentry.captureException()
- Add TypeScript types for error parameters

Users can now see what went wrong instead of errors being silently sent only to Slack, while still maintaining error tracking.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


henrikskog:
Ok. Closing this as I don't think it's worth fixing.

PR #2402, opened 2025-10-29
Change payment text

ninaflaaten:


junlarsen (approved):
Needs linting, otherwise LGTM

PR #2411, opened 2025-10-30
Add anniversary countdown banner

Kristiangit:


packages/config/tailwind.css
junlarsen: These changes should be put into web, otherwise there is no way to read the font iirc

junlarsen (changes requested):
Neat - I have some small comments, let me know what you think

junlarsen (approved):
Thanks

PR #2430, opened 2025-11-02
Fix attendees on waitlist not being reserved by task

andrefkl:
The task failed silently when `earliestReservationAt` was in the past which is always since the task is scheduled for that time


junlarsen (approved):

junlarsen:
> Should we even have a check for this in the task execution function?

It isnt strictly needed but I would assume the original intent was to be warned, just that the warning code was broken haha

PR #2436, opened 2025-11-05
Add `.gitattributes`

brage-andreas:
This might help terraform apply diffs on Windows to not be so huge


junlarsen (approved):
Do you want to do the same in https://github.com/dotkom/infra?

PR #2445, opened 2025-11-05
Create payments for users registered by deregistration

junlarsen:
Do not create payments for users on waitlist

Give payment links to users registered due to deregistration


apps/rpc/src/modules/event/attendance-service.ts
```diff
@@ -751,6 +751,22 @@ export function getAttendanceService(
...
+          attendee.paymentLink = payment.url
+          logger.info(
+            "Attendee(ID=%s,UserID=%s) has been reserved due to deregistration and given until %s UTC to pay for Event(ID=%s) at link %s",
+            attendee.id,
+            attendee.user.id,
+            paymentDeadline.toUTCString(),
+            event.id,
+            payment.url
+          )
+        }
```
andrefkl: Shouldn't this use firstUnreservedAdjacentAttendee?

brage-andreas:
[concerning](https://app.graphite.com/settings/meme-library?org=dotkom)

PR #2447, opened 2025-11-05
Show previous events in group page

sofieregine:


apps/web/src/app/grupper/components/GroupPage.tsx
```diff
@@ -16,22 +16,37 @@ export const GroupPage = async ({ params }: CommitteePageProps) => {
...
         orderBy: "asc",
       },
     }),
+    server.event.all.query({
+      filter: {
+        byEndDate: {
+          max: now,
+          min: null,
+        },
+        excludingOrganizingGroup: ["velkom"],
```
andrefkl: Maybe we should remove this in case the open group is velkom?

andrefkl (changes requested):
Nice! just one nit

andrefkl (approved):

PR #2449, opened 2025-11-05
Remove construction notice

jotjern:
The website is now perfect, so we do not need the construction notice anymore


junlarsen (approved):

brage-andreas:
Maybe also delete the component/file

PR #2475, opened 2025-11-09
Completely replace iconify with tabler icons

johannes-aas:


apps/web/src/app/arrangementer/components/AttendanceCard/MainPoolCard.tsx
```diff
@@ -98,7 +98,7 @@ export const MainPoolCard: FC<MainPoolCardProps> = ({ attendance, user, authoriz
           <div className="flex gap-[0.5ch] text-sm">
             <Text>Gå til</Text>
             <Link href="/profil" className="flex items-center text-blue-800 dark:text-blue-400 hover:underline">
-              <Text>profilsiden</Text> <Icon icon="tabler:arrow-up-right" className="text-base" />
+              <Text>profilsiden</Text> <IconArrowUpRight className="h-[1.25em] w-[1.25em]" />
```
junlarsen: You can set both at once with size https://tailwindcss.com/docs/width#setting-both-width-and-height

PR #2476, opened 2025-11-10
Add feedback tests

andrefkl:


junlarsen (commented):
Seems fine, perhaps we can rewrite some of the imperative loops using some of the nicer expect APIs? Worth to have a quick look imo https://vitest.dev/api/expect.html

junlarsen (approved):

PR #2500, opened 2025-11-14
Add external link icons to navbar

johannes-aas:
before:
image
after:
image

mobile:
image


apps/web/src/utils/is-link-external.ts
```diff
@@ -0,0 +1,3 @@
+export function isExternal(href: string) {
+  return href.startsWith("http://") || href.startsWith("https://")
```
junlarsen: Could argue we should check that domain is not online.ntnu.no

PR #2525, opened 2025-11-27
Standardize file uploading

brage-andreas:
This PR:

- (RPC) Creates a new route for each module for their own presigned url
    - Uses this new route instead of using Offline for everything
- (Dashboard) Updates ImageInput and FileInput to take a `onFileUpload` handler
- (Dashboard) Updates both inputs to be more user friendly
    - url field for image input
    - ability to remove input
- (Web) Updates usage in profile settings form
- Adds `created-by`​ metadata field that will be populated for every file uploaded to S3

### DIFF
--- a/apps/web/src/app/innstillinger/profil/form.tsx
+++ b/apps/web/src/app/innstillinger/profil/form.tsx
@@ -105,9 +105,9 @@ export function ProfileForm({ user, onSubmit, isSaving, saveSuccess, saveError, 
       return
     }
 
-    await trigger("imageUrl")
-
     onChange(result)
+
+    await trigger("imageUrl")
   }
 
   return (


apps/dashboard/src/components/forms/FileInput.tsx
```diff
@@ -1,32 +1,44 @@
...
-            placeholder={field.value ?? props.existingFileUrl ?? "Klikk for å velge fil"}
-            onChange={async (file) => {
-              if (file === null) {
-                return
-              }
-              const result = await upload(file)
-              field.onChange(result)
-            }}
-          />
+          <Stack gap="0.5rem">
```
junlarsen: Prefer to stick to mantine values, is `sm` too small?

PR #2544, opened 2025-11-29
Standardize `Article` repository and service layer

brage-andreas:


apps/rpc/src/modules/article/article-service.ts
```diff
@@ -69,48 +68,81 @@ export function getArticleService(
   s3Client: S3Client,
   s3BucketName: string
 ): ArticleService {
+  async function requireArticleExists(handle: DBHandle, articleId: ArticleId): Promise<boolean> {
+    const article = await articleRepository.findById(handle, articleId)
+
+    return Boolean(article)
+  }
```
junlarsen: you can drop this, and replace this with an `await getById()` at the callsite

apps/rpc/src/modules/article/article-service.ts
```diff
@@ -137,50 +169,63 @@ export function getArticleService(
...
         await articleTagRepository.delete(handle, tag)
       }
     },
+
     async setTags(handle, articleId, tags) {
-      const currentTags = (await articleTagRepository.getAllByArticle(handle, articleId)).map((tag) => tag.name)
-      const tagsToAdd = tags.filter((tag) => !currentTags.includes(tag))
-      const tagsToRemove = currentTags.filter((tag) => !tags.includes(tag))
-      const removePromises = tagsToRemove.map(async (tag) => this.removeTag(handle, articleId, tag))
+      await requireArticleExists(handle, articleId)
```
junlarsen: This piece of code does nothing, and does not ensure the article actually exists

PR #2569, opened 2025-12-01
Migrate `AuditLogRouter` to Authorization Middlewares

junlarsen:
Stacked onto ArticleRouter because ArticleRouter has fixes for propagated Context type

### DIFF
--- a/apps/rpc/src/modules/audit-log/audit-log-router.ts
+++ b/apps/rpc/src/modules/audit-log/audit-log-router.ts
@@ -4,11 +4,11 @@ import z from "zod"
 import { isAdministrator } from "../../authorization"
 import { withAuthentication, withAuthorization, withDatabaseTransaction } from "../../middlewares"
 import { BasePaginateInputSchema, PaginateInputSchema } from "../../query"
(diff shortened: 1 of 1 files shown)


apps/rpc/src/modules/audit-log/audit-log-router.ts
```diff
@@ -1,46 +1,76 @@
...
-    return ctx.executeTransaction(async (handle) => ctx.auditLogService.getById(handle, input))
-  }),
+export type GetAuditLogsByUserIdInput = inferProcedureInput<typeof getAuditLogsByUserIdProcedure>
+export type GetAuditLogsByUserIdOutput = inferProcedureOutput<typeof getAuditLogsByUserIdProcedure>
+const getAuditLogsByUserIdProcedure = staffProcedure
+  .input(PaginateInputSchema)
+  .use(withAuthentication())
+  .use(withAuthorization(isAdministrator()))
+  .use(withDatabaseTransaction())
+  .use(withAuditLogEntry())
```
brage-andreas: no

PR #2570, opened 2025-12-01
Migrate `CompanyRouter` to Authorization Middlewares

junlarsen:


apps/rpc/src/modules/company/company-router.ts
```diff
@@ -1,68 +1,101 @@
...
+    z.object({
+      filename: z.string(),
+      contentType: z.string(),
+    })
+  )
+  .output(z.custom<PresignedPost>())
+  .use(withAuthentication())
+  .use(withAuthorization(isEditor()))
+  .use(withDatabaseTransaction())
+  .use(withAuditLogEntry())
```
brage-andreas: this doesnt do anything as the route does not really use the database at all, but its fine

brage-andreas (approved):

PR #2591, opened 2025-12-02
Add isSameSubject authorization combinator

junlarsen:
Reads a UserId from the input and checks if they are both non-null and match the request principal.

### DIFF
--- a/apps/rpc/src/authorization.ts
+++ b/apps/rpc/src/authorization.ts
@@ -33,6 +33,7 @@
  * @packageDocumentation
  */
 
+import type { UserId } from "@dotkomonline/types"
 import { ADMIN_EDITOR_ROLES, type EditorRole } from "./modules/authorization-service"
 import type { Context, Principal } from "./trpc"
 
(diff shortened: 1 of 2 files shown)


apps/rpc/src/modules/user/user-router.ts
```diff
@@ -187,23 +199,21 @@ const updateUserProcedure = authenticatedProcedure
 export type IsStaffInput = inferProcedureInput<typeof isStaffProcedure>
 export type IsStaffOutput = inferProcedureOutput<typeof isStaffProcedure>
 const isStaffProcedure = procedure.use(withDatabaseTransaction()).query(async ({ ctx }) => {
```
brage-andreas: this is not part of the pr but why does this force using transactions?

brage-andreas (changes requested):

apps/rpc/src/authorization.ts
```diff
@@ -145,6 +146,26 @@ export function isGroupMember<TInput>(editorRole: EditorRole): Rule<TInput> {
   }
 }
 
+/**
+ * Business rule to check if the provided user ID (based on the procedure input) is equal to the principal performing
+ * the request.
+ *
+ * NOTE: A subject is the "correct" name for the ID of a user.
+ */
+export function isSameSubject<TInput>(
```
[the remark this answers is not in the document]
junlarsen: I agree isSameSubject isn't very descriptive, but I don't think isAuthorizedAs is any better. Authorization != Authentication

isSamePrincipal comes to mind, considering we use the term principal quite a lot in the code's documentation. isAuthenticatedAs maybe? isEqualToAuthenticatedUser?

apps/rpc/src/modules/user/user-router.ts
junlarsen: will be done in another PR

junlarsen:
Will merge, we can discuss naming later

PR #2598, opened 2025-12-02
Overload isGroupMember to accept EditorRole or Selector

junlarsen:


apps/rpc/src/authorization.ts
```diff
@@ -132,35 +132,51 @@ export function isEditor<TInput>(): Rule<TInput> {
...
+      const editorRole = selectorOrRole instanceof Function ? selectorOrRole(context.input) : selectorOrRole
+      if (editorRole === null) {
+        return false
+      }
       return context.principal.editorRoles.has(editorRole)
     },
   }
 }
 
+type IsSameSubjectSelector<TInput> = (input: TInput) => UserId | null
```
brage-andreas:
```suggestion
type SameSubjectSelector<TInput> = (input: TInput) => UserId | null
```

PR #2758, opened 2026-01-19
Fix deregistration not registering next attendee when capacity was previously full

andrefkl:


apps/rpc/src/modules/event/attendance-service.ts
```diff
@@ -809,7 +809,9 @@ export function getAttendanceService(
       const pool = attendance.pools.find((pool) => pool.id === attendee.attendancePoolId)
       invariant(pool !== undefined)
 
-      const attendeeCount = attendance.attendees.filter((a) => a.attendancePoolId === pool.id).length
+      const remainingAttendees = attendance.attendees.filter((a) => a.id !== attendee.id)
+
+      const attendeeCount = remainingAttendees.filter((a) => a.attendancePoolId === pool.id).length
       if (pool.capacity !== 0 && (pool.capacity < 0 || attendeeCount >= pool.capacity)) {
```
brage-andreas:
```suggestion
      if (pool.capacity !== 0 && (pool.capacity < 0 || attendeeCount > pool.capacity)) {
```
can't you just do this?

PR #2766, opened 2026-01-21
Add "Deployment" chapter to `CONTRIBUTING.md`

sondrealf:
Jeg brakk online.ntnu.no :P

Dette er litt docs for å sørge for at dette ikke skjer med nestemann. Det er bare å legge til kommentar eller comitts for å legge til mer info


brage-andreas:
Migrations should really be done by the release itself, but it's nice to document how to do it.

brage-andreas (approved):
I would probably also mention that only Dotkom members can do this, and is not relevant to any external party reading our code

PR #2771, opened 2026-01-22
Fix membership calculation

brage-andreas:
This PR fixes two bugs that were causing students to be assigned a start year 1 year later than reality (e.g., 2024 instead of 2023).

### Fix loop overwrite in empty semesters
The logic for estimating progress during empty semesters (like bachelor semester 4) was flawed. We iterated through all previous semesters to guess the current year, but we were overwriting the result in every iteration. Now we `Math#max` the current and local values to find the highest instead.

### Fix off-by-one rounding error
We were converting semesters to years using `Math.floor(semester / 2)`. This failed for spring semesters.

> Example: semester 3 (spring of 2nd year) became `floor(1.5) = 1`. This made the system think 2nd-year students were only in their 1st year.

This is fixed by changing it to `Math.round(semester / 2)`, so that semester 3 correctly resolves to 2 years.

### DIFF
--- a/apps/rpc/src/modules/user/user-service.ts
+++ b/apps/rpc/src/modules/user/user-service.ts
@@ -215,7 +215,7 @@ export function getUserService(
     },
 
     async register(handle, userId) {
-      // NOTE: The registrer function here has a few responsibilities because of our data strategy:
+      // NOTE: The register function here has a few responsibilities because of our data strategy:
       //
       // 1. The database is the source of truth, and is ALWAYS intended to be as such.
       // 2. Unfortunately, there was a period in time where Auth0 was the source of truth, most notably right after we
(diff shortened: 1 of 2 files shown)


brage-andreas:
* **#2772** <a href="https://app.graphite.com/github/pr/dotkom/monoweb/2772?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2771** <a href="https://app.graphite.com/github/pr/dotkom/monoweb/2771?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/dotkom/monoweb/2771?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

junlarsen (approved):

PR #2773, opened 2026-01-22
Add study grade calculation tests

brage-andreas:
I created a different PR for this because I don't know if this is even something we want due to how it changes with time, meaning the test data has to be in a specific time.


apps/rpc/src/modules/user/membership.e2e-spec.ts
```diff
@@ -847,4 +906,4 @@ const BACHELOR = {
       },
     ],
   },
-} as const satisfies Record<string, { goal: number; courses: NTNUGroup[] }>
+} as const satisfies Record<string, { goal: { year: number; grade: number }; courses: NTNUGroup[] }>
```
junlarsen:
```suggestion
} as const
```

junlarsen (approved):

PR #2774, opened 2026-01-23
Update membership system

brage-andreas:
Main changes:

- All `BACHELOR_STUDY` and `MASTER_STUDY` memberships now last exactly one semester
- Memberships are now semester-based instead of year-based
    - Logic has moved from date based to integer based
- The grade calculation is stored inside the membership
    - This makes it easier to adjust
- my chungus life


apps/rpc/src/modules/user/membership-service.ts
```diff
@@ -84,105 +119,132 @@ export function getMembershipService(): MembershipService {
...
-        let largestLocalSemester = 0
-        for (const previousSemester of previousSemesters.toReversed()) {
-          // The criteria here is the same as below, except we also require the course to have a finished date.
-          const hasPassedPreviousSemester = previousSemester.courses.every((course) =>
-            studentCourses.some(
-              (studentCourse) => course === studentCourse.code && studentCourse.finished !== undefined
-            )
-          )
+      // This semester has mandatory courses
+      if (semester.courses.length > 0) {
```
junlarsen:
```suggestion
      const hasMandatoryCourses = semester.courses.length > 0;
      if (hasMandatoryCourses) {
```

apps/rpc/src/modules/user/membership-service.ts
```diff
@@ -84,105 +119,132 @@ export function getMembershipService(): MembershipService {
...
+          studentCourses.some((studentCourse) => course === studentCourse.code && studentCourse.finished !== undefined)
+        )
+
+        if (coursesPassedEarlierSemester.length < earlierCourseSet.minimumEnrolledCourses) {
+          continue
+        }
+
+        const currentSemesterStart = isSpringSemester() ? getSpringSemesterStart() : getAutumnSemesterStart()
+
+        const earlierSemesterEstimates = earlierCourseSet.courses.map((course) => {
```
junlarsen: Needs some sort of explanation on what this huge map does

apps/rpc/src/scripts/migrate-users-from-ow4.ts
junlarsen: I think we can just delete these scripts? There is no OW4 API anymore or anything like that. That honestly goes for most if not all import scripts

packages/utils/src/semester-helpers.ts
```diff
@@ -0,0 +1,139 @@
...
+  const year = typeof dateOrYear === "number" ? dateOrYear : dateOrYear.getFullYear()
+
+  return new TZDate(year, JANUARY, 1)
+}
+
+/**
+ * Get the start of the academic year, which is by our convention August 1st.
+ * `August 1st -- <year>-08-01T00:00:00.000Z`
+ */
+export const getAutumnSemesterStart = (dateOrYear: TZDate | Date | number = getCurrentUTC()) => {
```
junlarsen: Same here

packages/utils/src/semester-helpers.ts
```diff
@@ -0,0 +1,139 @@
...
+  return semesterDifference
+}
+
+export function getStudyGrade(semester: number): number
+export function getStudyGrade(semester: null | undefined): null
+export function getStudyGrade(semester: number | null | undefined): number | null
+export function getStudyGrade(semester: number | null | undefined): number | null {
+  if (!semester) {
+    return null
+  }
```
junlarsen: Why should we ever accept that semester is null or undefined here? It should be the callers responsibility.

apps/rpc/src/scripts/migrate-users-from-ow4.ts
brage-andreas: I think so too

packages/utils/src/semester-helpers.ts
```diff
@@ -0,0 +1,139 @@
...
+import { getCurrentUTC } from "./utc"
+
+const JANUARY = 0
+const AUGUST = 7
+
+/**
+ * Get the start of the academic year, which is by our convention January 1st.
+ * `January 1st -- <year>-01-01T00:00:00.000Z`
+ */
+export const getSpringSemesterStart = (dateOrYear: TZDate | Date | number = getCurrentUTC()) => {
```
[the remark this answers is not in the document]
brage-andreas: We don't. I did originally but I think it's nice that the helpers are nice to use. Most of the helpers added aren't used (I added more since you reviewed), but I think there is value of just having them available so the next person don't have to dive into this

packages/utils/src/semester-helpers.ts
brage-andreas: I think it's nice for a helper to allow inputs to be nullish. Regardless, I think it will help with avoiding cases like `const grade = membership?.semester ? getStudyGrade(membership.semester) : null`

brage-andreas:
Afaik this is finished

packages/utils/src/semester-helpers.ts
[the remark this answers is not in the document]
junlarsen: What I am asking if there is any reason to accept number here? For what it's worth, both Date and number are not timezone aware so we should prefer to not use them

packages/utils/src/semester-helpers.ts
```diff
@@ -0,0 +1,179 @@
+import { TZDate } from "@date-fns/tz"
+import { addYears, isBefore, subYears } from "date-fns"
+import { getCurrentUTC } from "./utc"
+
+const JANUARY = 0
+const AUGUST = 7
```
junlarsen: God bless javascript 🙃

PR #2825, opened 2026-01-31
Script: Vinstraff user db sync

sondrealf:
It's very hard to be able to run Vinstraff reliably local. As to run Vinstraff locally, you need to run monoweb locally to give vinstraff user authentication. 

This script syncs the comitee, users and group relations from prod to local, so that nessecary data is in place. This could also be used for Monoweb user testing to create a more realistic testing env.

Have already updated readme to [Vinstraff](https://github.com/dotkom/vengeful-vineyard). Will also update it again after this pr.

### DIFF
--- a/apps/brevduen/package.json
+++ b/apps/brevduen/package.json
@@ -22,7 +22,7 @@
     "zod": "^3.25.47"
   },
   "devDependencies": {
-    "@biomejs/biome": "2.3.11",
+    "@biomejs/biome": "2.3.13",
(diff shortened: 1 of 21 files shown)


brage-andreas:
- Im not able to read it over yet, but users won't have the same id as in prod. Is that taken care of?

sondrealf:
> * Im not able to read it over yet, but users won't have the same id as in prod. Is that taken care of?

Yes this is now taken care off

PR #2858, opened 2026-02-03
Only show events where user is registered

junlarsen:


andrefkl (commented):
This is used by a couple other things like the dashboard. I think we should either add a query filter or a different method/route for events the user is reserved on

PR #2863, opened 2026-02-04
Use generated pool title if it was not previously generated

andrefkl:
A lot of people seems to create a pool with all year criteria (1-5), and later remove some criteria, but the title stays as "Alle".

This pr should make pools use generated titles as long as:
1. It wasn't set to a custom value when previously created/updated
2. It hasn't been manually edited this session

### DIFF
--- a/apps/dashboard/src/app/(internal)/arrangementer/components/pool-form.tsx
+++ b/apps/dashboard/src/app/(internal)/arrangementer/components/pool-form.tsx
@@ -63,8 +63,8 @@ export const usePoolForm = (props: PoolFormProps) => {
 
   const generatedTitle = createPoolName(yearCriteria ?? [])
   const defaultTitle = form.formState.defaultValues?.title
-  const defaultTitleWasGenerated = defaultTitle === createPoolName(props.defaultValues.yearCriteria ?? [])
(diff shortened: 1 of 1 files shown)


apps/dashboard/src/app/(internal)/arrangementer/components/pool-form.tsx
```diff
@@ -63,10 +63,8 @@ export const usePoolForm = (props: PoolFormProps) => {
 
   const generatedTitle = createPoolName(yearCriteria ?? [])
   const defaultTitle = form.formState.defaultValues?.title
-
-  const titleIsDirty =
-    Boolean(form.formState.dirtyFields.title) ||
-    (form.getValues("title") === defaultTitle && defaultTitle !== generatedTitle && defaultTitle !== "")
+  const defaultTitleWasGenerated = defaultTitle === createPoolName(props.defaultValues.yearCriteria ?? [])
+  const titleIsDirty = Boolean(form.formState.dirtyFields.title)
```
junlarsen: nit: naming


```suggestion
  const isDefaultGeneratedTitle = defaultTitle === createPoolName(props.defaultValues.yearCriteria ?? [])
  const isTitleDirty = Boolean(form.formState.dirtyFields.title)
```

PR #2913, opened 2026-02-17
Add max file size frontend validation

andrefkl:
* Move max file sizes to consts in types
* Use separate s3 presigned post upload method for offline images and files
* Show max file size as input description and form error
* Stop throwing `No location header` error when s3 upload fails


brage-andreas (approved):
Looks amazing!

PR #2939, opened 2026-02-21
Use summary types for event list to speedup queries

andrefkl:
* Adds indexes on heavy join tables
* Uses EventSummary and AttendanceSummary for event lists that doesn't include attendees or other unnecessary data
* Removes some unnecessary calls that slows down pages


brage-andreas (approved):
Nice! Looks good

PR #2953, opened 2026-02-25
Add navbar and footer

sofieregine:
Added footer and navbar. Also added backgroundcolor, both for dark and lightmode.


ellawaal (changes requested):
Nice work! I have commented some changes I would like to see:)

PR #2982, opened 2026-03-04
Add nuqs and tabs to switch between events user was reserved and in waitlist for in profile page

andrefkl:
image

We can use nuqs later to clean up the event filter states etc and add more state to the url in other places

### DIFF
--- a/apps/web/src/app/profil/[profileSlug]/ProfilePage.tsx
+++ b/apps/web/src/app/profil/[profileSlug]/ProfilePage.tsx
@@ -9,18 +9,19 @@ import { useTRPC } from "@/utils/trpc/client"
 import { useFullPathname } from "@/utils/use-full-pathname"
 import { useSession } from "@dotkomonline/oauth2/react"
 import {
-  type Membership,
-  type VisiblePersonalMarkDetails,
   createGroupPageUrl,
   findActiveMembership,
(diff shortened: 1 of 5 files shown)


junlarsen (changes requested):
Keep the url state as a zod schema preferably https://nuqs.dev/docs/parsers/community/zod-codecs

I do however think that this isn't really the right direction to do form-url syncing. I can show you tomorrow @andrefkl

apps/web/src/app/profil/[profileSlug]/ProfilePage.tsx
```diff
@@ -254,6 +293,19 @@ export function ProfilePage() {
...
+
+  const eventTabsForm = useForm<EventListTabFormSchema>({
+    defaultValues: {
+      eventListTab: eventListTabQueryParam,
+    },
+  })
+
+  const submitEventTabsForm = eventTabsForm.handleSubmit(({ eventListTab }) => {
+    router.replace(`?eventListTab=${eventListTab}`, { scroll: false })
+  })
```
Comment (junlarsen) [inline apps/web/src/app/profil/[profileSlug]/ProfilePage.tsx]: Better way to handle this is to make the form not submittable, and instead just do it on change with subscribe()

https://react-hook-form.com/docs/useform/subscribe

junlarsen (changes requested):

PR #3077, opened 2026-04-05
Solve UI library react/typescript performance issues

jotjern:


packages/ui/src/atoms/Button/Button.tsx
```diff
@@ -12,21 +10,23 @@ export type ButtonVariant = (typeof BUTTON_VARIANTS)[number]
...
-export type ButtonProps<E extends ElementType = "button"> = VariantProps<typeof button> &
-  PropsWithChildren & {
-    /**
-     * The HTML element to render the button as
-     *
-     * Defaults to an HTML <button> element, but can be used with the Link
-     * component from 'next/link' to create a link that looks like a button
-     */
-    element?: E
+export type ButtonProps = VariantProps<typeof button> &
```
junlarsen, line 13: This breaks the types for polymorphic components so I don't quite know if we should do this

PR #3078, opened 2026-04-08
Use custom ranking for courses when searching

henrygraesberg:


apps/grades-backend/src/modules/course/course-repository.ts
```diff
@@ -13,60 +13,81 @@ export interface CourseRepository {
...
+          "pass_rate" AS "passRate",
+          "created_at" AS "createdAt",
+          "updated_at" AS "updatedAt",
+          "taught_semesters" AS "taughtSemesters",
+          "teaching_languages" AS "teachingLanguages",
+          "campuses",
+          "faculty_id" AS "facultyId",
+          "department_id" AS "departmentId"
+        FROM "course"
+        WHERE 1 = 1
```
andrefkl: Why?

andrefkl (approved):

junlarsen (commented):
Would it maybe be more ergonomic to define the ranking function as a postgresql procedure?

andrefkl (approved):

junlarsen (approved):

packages/grades-db/prisma/migrations/20260428154840_add_course_ranking_function/migration.sql
```diff
@@ -0,0 +1,38 @@
...
+    search_starts_with TEXT;
+    search_contains TEXT;
+    baseline INTEGER;
+BEGIN
+    IF search_term IS NULL THEN
+        RETURN 0;
+    END IF;
+
+    search_starts_with := search_term || '%';
+    search_contains := '%' || search_term || '%';
```
junlarsen, line 20: Just a heads up, if you use `%` as the first character of a LIKE/ILIKE, postgresql cannot use any index on the column. This likely results in a full table scan.

PR #3517, opened 2026-08-18
Update time location section in event page

brage-andreas:

PR #3458, opened 2026-08-11
Add event cards

brage-andreas:

PR #3518, opened 2026-08-18
Make page filters persistent

brage-andreas:


brage-andreas:
* **#3518** <a href="https://app.graphite.com/github/pr/dotkom/monoweb/3518?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/dotkom/monoweb/3518?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3477** <a href="https://app.graphite.com/github/pr/dotkom/monoweb/3477?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3458** <a href="https://app.graphite.com/github/pr/dotkom/monoweb/3458?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#3517](https://github.com/dotkom/monoweb/pull/3517) <a href="https://app.graphite.com/github/pr/dotkom/monoweb/3517?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#3516** <a href="https://app.graphite.com/github/pr/dotkom/monoweb/3516?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/utils/src/persistent-filters.ts
andrefkl, line 1: I think we should use cookies instead of localStorage so we don't get a layout shift from the server to client

apps/dashboard/src/app/(internal)/arrangementer/hooks/use-event-filters.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    router.replace(`/arrangementer?${params.toString()}`, { scroll: false })
+  }, [filters, router, searchParams])
+
+  const updateFilters = useCallback(
+    (partial: Partial<EventFilters>) => {
+      const nextFilters = { ...filters, ...partial }
+      const params = new URLSearchParams(searchParams.toString())
+
+      persistFilters(EVENT_FILTERS_STORAGE_KEY, nextFilters, isDefaultFilters)
```
andrefkl: Imo we should only persist things that change how the user views the data, i.e. sort chronological vs grouped, and view events as cards vs a list. Not filters or search

apps/dashboard/src/app/(internal)/arrangementer/hooks/use-event-filters.ts
andrefkl, line 1: I think using nuqs for this would make this a lot simpler. Then we wouldn't need a use[domain]Filters file for every page we want filters.

apps/web/src/app/artikler/ArticleFilters.tsx
andrefkl, line 1: Found a bug here. If you go to /artikler and search for something, then remove the `q`param in the url and refresh then the whole thing glitches. The same thing happens in /karriere 

[Screen Recording 2026-08-19 at 16.39.26.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/128951ed-9ea7-4241-a688-08c68b61e29d.mov)

andrefkl (approved):
I still don't think we should use localStorage for the view but the rest looks fine

PR #3562, opened 2026-08-31
Update event page navigation

johannes-aas:
This PR makes the icon buttons wider for small screens making them easier to press. It collapses into "icon-only" sooner on calendar view, to avoid wrapping. Also replaced the card layout icon with a more symmetrical one.

Before:

image

image

After:

image

image


brage-andreas (approved):
Nice!

PR #3643, opened 2026-09-09
Add a tab that displays active and inactive groups that the user is or has been in 

tharyng:
Screenshot 2026-09-09 204404
Screenshot 2024-09-16 144007


andrefkl (approved):
🤯

PR #3644, opened 2026-09-09
Add file size warning to offline file upload

andrefkl:

### DIFF
--- a/apps/dashboard/src/app/(internal)/offline/modals/create-offline-modal.tsx
+++ b/apps/dashboard/src/app/(internal)/offline/modals/create-offline-modal.tsx
@@ -1,4 +1,5 @@
 import type { OfflineWrite } from "@dotkomonline/rpc/offline"
+import { getCurrentUTC } from "@dotkomonline/utils"
 import { type ContextModalProps, modals } from "@mantine/modals"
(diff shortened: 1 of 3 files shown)


andrefkl:
* **#3644** <a href="https://app.graphite.com/github/pr/dotkom/monoweb/3644?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/dotkom/monoweb/3644?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

brage-andreas (commented):
i think the huge red button is kind of invisible. it also shouldn't be red. i think this would make more sense as an alert modal of some sort.

andrefkl:
> i think the huge red button is kind of invisible. it also shouldn't be red. i think this would make more sense as an alert modal of some sort.

​Can't really make it into a modal cause then we have stacked modals which doesn't really work in the ui. But i can change the button color

brage-andreas:
I think it should replace the submit button in that case. "Last opp stor fil likevel", then the current submit button

andrefkl (commented):
Closing this as it's fixed in #3699

PR #3654, opened 2026-09-11
Add bug report form in profile menu and contact section in footer

### LINKED ISSUE #3552: Add bug report button
With a form they can write in and submit which sends an email to us

johron:
Added form with two text inputs/areas for subject and body which can then be sent with mailto

Some minor things (TODO):
* This is using ´Section´ from ´bedrift/faktura/components/section´, which is a component meant for something else

### DIFF
--- a/apps/web/src/components/BugReportForm.tsx
+++ b/apps/web/src/components/BugReportForm.tsx
@@ -4,40 +4,23 @@ import {
   AlertDialogCancel,
   AlertDialogContent,
   AlertDialogTitle,
-  AlertDialogTrigger,
   Button,
   Text,
   Textarea,
   TextInput,
   Title,
 } from "@dotkomonline/ui"
-import { IconMail, IconX } from "@tabler/icons-react"
+import { IconX } from "@tabler/icons-react"
 import { Section } from "@/app/bedrift/faktura/components/section"
 
-interface BugReportButtonProps {
+export interface BugReportFormProps {
   bugReportFormOpen: boolean
   setBugReportFormOpen: (open: boolean) => void
 }
 
-export const BugReportButton = ({ bugReportFormOpen, setBugReportFormOpen }: BugReportButtonProps) => {
+export const BugReportForm = ({ bugReportFormOpen, setBugReportFormOpen }: BugReportFormProps) => {
   const [title, setTitle] = useState("")
   const [description, setDescription] = useState("")
-
-  const button = (
-    <button
-      type="button"
-      aria-label="Send bugreport via e-post"
-      className="group relative flex flex-row items-center gap-2 cursor-pointer hover:bg-gray-100 dark:hover:bg-stone-800 px-1 -mx-1 py-0.25 -my-0.25 rounded-md hover:-ml-6 hover:pl-6"
-    >
-      <Text>Ta kontakt med Dotkom</Text>
-      <div
-        className="pointer-events-none absolute top-1/2 left-1 -translate-y-1/2 invisible group-hover:visible"
-        aria-hidden
-      >
-        <IconMail className="shrink-0 size-4" />
-      </div>
-    </button>
-  )
 
   const sendEmail = () => {
     const subject = encodeURIComponent(title)
@@ -47,7 +30,6 @@ export const BugReportButton = ({ bugReportFormOpen, setBugReportFormOpen }: Bug
 
   return (
     <AlertDialog open={bugReportFormOpen} onOpenChange={setBugReportFormOpen}>
-      <AlertDialogTrigger asChild>{button}</AlertDialogTrigger>
       <AlertDialogContent size="lg" className="p-0!" onOutsideClick={() => setBugReportFormOpen(false)}>
         <div className="flex items-center justify-between px-4 pt-4 rounded-t-lg">
           <AlertDialogTitle asChild>
--- a/apps/web/src/components/Footer/ContactSection.tsx
+++ b/apps/web/src/components/Footer/ContactSection.tsx
@@ -20,6 +20,11 @@ export const ContactSection = () => {
         </Text>
 
         <div>
+          <Text>Oppdaget en feil på nettsiden?</Text>
+          <BugReportButton bugReportFormOpen={bugReportFormOpen} setBugReportFormOpen={setBugReportFormOpen} />
+        </div>
+
+        <div>
           <Text>Har du lyst til å bidra?</Text>
           <Text>
             Kildekoden finner du på{" "}
@@ -27,17 +32,6 @@ export const ContactSection = () => {
               Github
             </a>
           </Text>
-        </div>
-
-        <div>
-          <Text>Oppdaget en feil på nettsiden?</Text>
-          <BugReportButton bugReportFormOpen={bugReportFormOpen} setBugReportFormOpen={setBugReportFormOpen} />
-          {/*<Text>
-            Ta kontakt med{" "}
-            <a className="underline" href="mailto:dotkom@online.ntnu.no" target="_blank" rel="noopener noreferrer">
-              Dotkom
-            </a>
-          </Text>*/}
         </div>
       </div>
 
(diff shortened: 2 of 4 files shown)


apps/web/src/components/BugReportForm.tsx
```diff
@@ -0,0 +1,78 @@
...
+  AlertDialogContent,
+  AlertDialogTitle,
+  Button,
+  Text,
+  Textarea,
+  TextInput,
+  Title,
+} from "@dotkomonline/ui"
+import { IconX } from "@tabler/icons-react"
+import { Section } from "@/app/bedrift/faktura/components/section"
```
andrefkl: You don't need to use this, you can just use normal html tags with className="flex flex-col gap-4"

apps/web/src/components/BugReportForm.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+export const BugReportForm = ({ bugReportFormOpen, setBugReportFormOpen }: BugReportFormProps) => {
+  const [title, setTitle] = useState("")
+  const [description, setDescription] = useState("")
+
+  const sendEmail = () => {
+    const subject = encodeURIComponent(title)
+    const body = encodeURIComponent(description)
+    window.open(`mailto:dotkom@online.ntnu.no?subject=${subject}&body=${body}`, "_blank", "noopener,noreferrer")
+  }
```
andrefkl: Use react-hook-form here instead, and call `sendEmail` in `onSubmit` on the form. You can see an example of how to use it in https://github.com/dotkom/monoweb/blob/main/apps/web/src/app/arrangementer/components/DeregisterModal.tsx#L100

apps/web/src/components/BugReportForm.tsx
```diff
@@ -0,0 +1,78 @@
...
+  Textarea,
+  TextInput,
+  Title,
+} from "@dotkomonline/ui"
+import { IconX } from "@tabler/icons-react"
+import { Section } from "@/app/bedrift/faktura/components/section"
+
+export interface BugReportFormProps {
+  bugReportFormOpen: boolean
+  setBugReportFormOpen: (open: boolean) => void
```
andrefkl:
```suggestion
  isBugReportFormOpen: boolean
  setIsBugReportFormOpen: (open: boolean) => void
```

We usually prefix booleans with `is` or similar

apps/web/src/components/BugReportForm.tsx
```diff
@@ -0,0 +1,78 @@
...
+        </div>
+        <div className="flex flex-col gap-1 px-4 pb-4 rounded-lg min-h-[25dvh] max-h-[75dvh] overflow-y-auto">
+          <div className="flex flex-col gap-2">
+            <Section as="fieldset" className="gap-8">
+              <Section>
+                <Text>Beskriv feilen kort og tydelig</Text>
+                <Section>
+                  <TextInput
+                    label="Kort tittel"
+                    placeholder="F.eks. problem med ..."
```
andrefkl: I think its nice if we give a concrete example

```suggestion
                    placeholder="Jeg får ikke meldt meg på et arrangement"
```

apps/web/src/components/BugReportForm.tsx
```diff
@@ -0,0 +1,78 @@
...
+                    required={true}
+                    value={title}
+                    onChange={(event) => setTitle(event.currentTarget.value)}
+                  />
+                </Section>
+                <Section>
+                  <Title element="legend">Hva skjedde?</Title>
+                  <Textarea
+                    label="Fortell hva du gjorde, hva du forventet og hva som faktisk skjedde"
+                    placeholder="F.eks. når man gjør ditt og datt så ..."
```
andrefkl:
```suggestion
                    placeholder="Jeg prøvde å melde meg på ITEX, men fikk feilmeldingen..."
```

apps/web/src/components/BugReportForm.tsx
```diff
@@ -0,0 +1,78 @@
...
+                    required={true}
+                    value={description}
+                    onChange={(event) => setDescription(event.currentTarget.value)}
+                  />
+                </Section>
+                <Button size="lg" variant="default" type="submit" onClick={sendEmail}>
+                  Send feilen til oss
```
andrefkl:
```suggestion
                  Send
```

apps/web/src/components/Footer/ContactSection.tsx
```diff
@@ -2,115 +2,130 @@
...
-import { IconCheck, IconClipboard } from "@tabler/icons-react"
+import { IconCheck, IconClipboard, IconMail } from "@tabler/icons-react"
+import { useState } from "react"
+import { BugReportForm } from "../BugReportForm"
 
 export const ContactSection = () => {
   const { icon: copyKontaktEmailIcon, copy: copyKontaktEmail } = useCopyToClipboard()
   const { icon: copyOkonomiEmailIcon, copy: copyOkonomiEmail } = useCopyToClipboard()
 
+  const [bugReportFormOpen, setBugReportFormOpen] = useState(false)
```
andrefkl:
```suggestion
  const [isBugReportFormOpen, setIsBugReportFormOpen] = useState(false)
```

apps/web/src/components/Navbar/ProfileMenu.tsx
```diff
@@ -239,6 +195,52 @@ export const AvatarDropdown: FC<AvatarDropdownProps> = ({ dbUser }) => {
...
+          icon: IconMailForward,
+          label: "Kontakt oss",
+          href: "mailto:hovedstyret@online.ntnu.no",
+          openInNewTab: true,
+        },
+        {
+          icon: IconBug,
+          label: "Rapporter en feil",
+          onClick: () => setBugReportFormOpen(true),
+        },
```
andrefkl: I think its cleaner to not include `Rapporter en feil` in `linkGroups`, and instead render it after `filteredLinkGroups`, as it also shouldn't be a Link with onClick. Then you can just render it as a `DropdownMenuItem` with an onClick directly.

andrefkl (changes requested):
Nice work! A few comments, most are just to show the conventions used in the repo.

I think the main goal of this feature should be to remove the friction of users sending emails themselves, and that we should send the email via the backend on submit instead of using `mail:to`.

Suggested flow:

- If the user is logged in: Include a checkbox "Send anonymt" in the form.
- If they're not logged in: Include a text input "E-post (valgfritt, hvis du vil ha svar).

If the user is not anonymous, set `Reply-To` on the email to their email, and potentially include their name and email at the top of the email body.

johron:
Okay. I'll try to get those things implemented!

johron:
I do not know why copilot activated? I May have pressed a wrong Button.. I'll try to remove that commit, because it conflicts with local changes

johron:
> I think the main goal of this feature should be to remove the friction of users sending emails themselves, and that we should send the email via the backend on submit instead of using `mail:to`.

How would emailing go, would we be sending the email to ourselves? Also I'm not entirely sure how to send emails through the backend, I've been trying to walk through the code and I've found `email-service.ts` with `new SendEmailCommand` and `sesClient.send(sendEmailCommand)` and `getEmailService` with `send` function.

Also how does emails work when running locally, does it still work, are there any differences?

andrefkl:
> > I think the main goal of this feature should be to remove the friction of users sending emails themselves, and that we should send the email via the backend on submit instead of using `mail:to`.
> 
> How would emailing go, would we be sending the email to ourselves? Also I'm not entirely sure how to send emails through the backend, I've been trying to walk through the code and I've found `email-service.ts` with `new SendEmailCommand` and `sesClient.send(sendEmailCommand)` and `getEmailService` with `send` function.
> 
> Also how does emails work when running locally, does it still work, are there any differences?

Yes, the backend would call EmailService and we would send the email ourselves. But there's a lot of extra things to it and it's a bit complicated, especially for a first issue. We can look at it together on wednesday if you want?

Emails are turned off in dev, so you can call EmailService but nothing would happen.

johron:
Oh okay. I'm down to work on it then, if possible considering introduction stuff, I don't know the plan for Wednesday.

